PR #2, opened 2015-12-25
LocalDate.minusDays() implementation and test enhancements

phueper:


phueper:
Hey @pithu ... just out of curiosity ... care for some contributions?

lbilharz:
:+1:

phueper:
Argh... i forgot that github is adding all commits to an open pull request on a branch... oh well... so i played around with travis and got rid of  firefox on travis instead using phantomjs... you might wanna reject this PR and i can create a new one if you like ;)

pithu:
contributions, yes great !!!! thx a lot patty !

but puh, so many different apsects in one pull request.

1) minusDays
my idea is to stay as close as possible to the openjdk 8 implementation (JSR310) for mainly two reason:
- if you know the JSR310 API you should feel comfortable with js-joda API
- minimize unexpected bugs or missbehaviour
At http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/time/LocalDate.java#l1486, they made it a bit different. Do you know why ? Do we need that?

2) minusDays tests
Do we really need this bunch of tests ? I would like to add only really necessary tests. 
I am not really happy with the tests right now. I would like to just reuse the jdk tests or at least use the same test cases. Did you had a look on that ?

3 ) node 10
using node 10 was by purpose. Why not being downward compatible to an older node version? Will be there anything special we will need of the node 4 features. (i hope not ;) )

4) Phantom vs Firefox
i am unsure with this point. Since travis is doing the job, we could even add both browser to the test job. But is this really necessary. Since we are independent from any Dom feature, and using babel to transpile es5 code out of the es6 code, i do not really expect any issues here. 

But on the other side, me, as a developer want to run all travis tests before pushing to the main line. So i would have to install phantomjs locally. And, sorry, i had so much trouble with phantomjs ... 
But the same might go for Firefox. Not every developer like this browser and is not willing to install it locally

Hey patty, can you split the pull request it separate peaces, so we can discuss/ merge it separately.

Especially at the beginning of this project its quite important to find the right vision and philosophy for the right pass for this project.

P.S. and whats about Christmas ???   ;)

phueper:
Yeah... I think you are right... Way too many aspects...  I'll close it and create separate PRs

PR #4, opened 2015-12-26
Tests for negative arguments to plusDays and minusDays

phueper:
Base for this is PR #3

I think if we do accept negative arguments we should also test them.

On the other hand i like the idea to use the JDK Tests as a kind of "reference" and reuse them or write tests that are based on the JDK Tests... so this PR is more for discussion and completeness sake... 
The question is how you would like to handle this... should we rather create an issue to reuse the JDK Tests as reference base?

### DIFF
--- a/src/LocalDate.js
+++ b/src/LocalDate.js
@@ -111,7 +111,7 @@ export class LocalDate {
      * @throws AssertionError if the result exceeds the supported date range
      */
     minusDays(daysToSubtract) {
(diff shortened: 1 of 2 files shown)


phueper:
so... should we take the TCK Tests from http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/test/java/time/tck/java/time and port them to mocha/chai?

PR #6, opened 2015-12-27
TCK tests proposal

phueper:
i transformed one of the tests from the TCKLocalDate.java class


phueper:
I have more of these `of` factory tests coming up... maybe we can take a look at them when there are more tests... maybe i can finish them tonight

PR #7, opened 2015-12-30
TCK Tests

phueper:
- more functionality added to `Month` and `LocalDate`
- all Tests in `tck/LocalDateTest.js` now enabled


pithu:
ups sorry, changed Month as well in the same time. I changed the way to export "static const class properties". Like that it feels a bit more handy, doesnt it ?
```
import {Month} from './Month'
Month.of(1);
Month.JANUARY;
```
Dont we need a validation for the monthOfYear range?

phueper:
Hmm... you are right... i thought i took that from the JDK... but it is not in there... i'll redo it :/

phueper:
done with the `of...` tests :)

PR #14, opened 2016-01-13
update README.md

### LINKED ISSUE #10: License
@pithu, @phueper -- I think it would be very helpful for this project if the two of you could agree on a license for this code and officially add it to the repo.

You've gotten a little attention recently in discussion at https://github.com/moment/moment/issues/1754#issuecomment-169142952, and I suspect you might pick up another contributor or two from that discussion if your licensing intentions were a bit clearer.  (I might be interested to contribute myself at some point, but I'm pretty swamped for at least the next month or two.)

My personal bias is that the original Joda Time got this right when they picked the Apache License -- but as the two folks who have submitted code, obviously your preferences reign supreme.

Thanks for working on this library!

### LINKED ISSUE #13: Why JodaTime; why not java.time?
As you probably know, Joda-Time influenced JSR-310, which became `java.time` in Java 8.  Several things were changed along the way, as described by Stephen Colebourne here: http://blog.joda.org/2009/11/why-jsr-310-isn-joda-time_4941.html

So - is this a Joda Time port to JavaScript, or a java.time port to JavaScript?   If the former, then *why*, considering the improvements and the young state of this port.  If the latter, then the description in the readme and project page should change.

FYI - Much of my experience comes from contributing to Noda Time, which is more like java.time than it is with Joda.  I'm also a collaborator on moment.js, but I've been toying with the idea of a Noda Time or java.time port to JavaScript for awhile now, so I'd be happy to help where I can.

Time zones are going to be the biggest challenge. :)

phueper:
- ThreeTen-Backport reference
- License
- Roadmap

fixes #10, 
fixes #13

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,33 @@
+```
+/*
+ * Copyright (c) 2016, Philipp Thuerwaechter & Pattrick Hueper
+ *  
+ * All rights reserved.
+ *  
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *  
+ *  * Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *  
+ *  * Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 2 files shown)


pithu:
Perfect, thx patty

PR #16, opened 2016-01-20
Monster PR (Month, Duration, ThreeTenBP, ...)

phueper:
Sorry... this has gotten way too big

- started as the Month/MonthTests
- i realized i needed Duration as well so started the DurationTests
- switched to the ThreeTenBP as reference during the licensing discussion
- finalized the DurationTests

i doubt it is feasible to break this apart into more "digestible" chunks... so we'll probably merge it as a whole to have a base for further, smaller PRs


phueper:
oh well... i guess i need to increase the coverage some more :/

PR #25, opened 2016-03-19
cleanup TODOs in Duration and DurationTest

phueper:
- the overflow in toSeconds() is not acceptable and not intended, since Duration is calculating seconds as floats internally...
  we use normal multiplication here and accept the possible rounding error
- the rounding errors in dividedBy are also acceptable since we use normal float division as well
(cherry picked from commit cc4840b)


timaschew:
Why not using a library like [big.js](https://github.com/MikeMcl/big.js)?
I used it for the Duration class: https://github.com/Kenobi-Time/KenobiTime/blob/525f4ac8229f17748f1b0202d9b02d343c6dbab8/lib/time/duration.coffee#L22-L26

pithu:
IMHO floating point divisions are not acceptable. We should solve issues like this, by converting to a int div and int mod. 

`d.mulltipliedBy(2).equals(d.plus(d))` should be true  always!

The following (untested pseudo) code should work, doing a ROUND_DOWN not HALF_DOWN and might have an overflow for a multiplicand bigger then 9.007.199. Both is acceptable for me.

 ```
    multipliedBy(multiplicand) {
        var secs = this._seconds * multiplicand;
        var nos = this._nanos * multiplicand;
        if (nos > MAX_SAFE_INTEGER || nos < MIN_SAFE_INTEGER){
            // overflow exceptions
        }
        secs = secs + MathUtil.intDiv(nos, LocalTime.NANOS_PER_SECOND);
        nos = MathUtil.intMod(nos, LocalTime.NANOS_PER_SECOND);
        return Duration.ofSeconds(secs, nos);
    }
```

phueper:
@pithu: I updated `multipliedBy` with your proposal slightly modified to use `MathUtil.safeMultiply` ... that works, however the rounding differences remain

pithu:
:+1:

pithu:
@timaschew , as a basic, core library,  we don't want to have any third party dependencies.

PR #28, opened 2016-03-20
Simplify deployment for gh-pages

timaschew:
This is one of two parts to simplify the deployment.

With this setup you only need to sync the two markdown files (readme and cheat-sheet) from the master branch. If you change them just run `npm run gh-pages`, that's it.

### DIFF
--- a/deploy-gh-pages.js
+++ b/deploy-gh-pages.js
@@ -1,0 +1,15 @@
+var ghpages = require('gh-pages');
+var path = require('path');
+
+ghpages.publish(path.join(__dirname, '.publish'), {
+  src: ['_includes/README.md', '_includes/CheatSheet.md'],
+  branch: 'gh-pages',
+  add: true, // otherwise all other files will be deleted
+  message: 'Auto-generated commit'
+}, function(err, res) {
+  if (err != null) {
+    console.log(err);
+  } else {
+    console.log('pushed gh-branch');
+  }
+});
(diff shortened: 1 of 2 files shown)


pithu:
that is a lot of code for a simple git merge. i prefer to do it manually. 
Please let us follow the golden rule, no extra code wrapper around "standard work flows / tools"

timaschew:
It's a git merge from one branch to another branch but changing the directory.
Changing the directory is not supported by the __gh-page__ module.
There is another module for gulp, which allow easily to modify the filenames and directories, but it doesn't support the `add` option, which results to an effect which deletes all other files.
So this task is not so simple as it sounds at the first glance.

I would like to know which commands do you execute exactly when doing this manually.

pithu:
OK, its not a simple merge, but its a copy or cherry pick. And the files differs slightly anyway, a merge would have conflicts. But this is not my point. 
 
It is very important to me, to keep the structure of js-joda simple, clear and unified.

Therefore we have to agree on some rules, in my opinion that are things like:
- use unified pattern
- only use standard config files
- do not add any special script files
- put things together that belongs together
- etc.

In my opinion your pull request do not align with the rules because: 
- all build output should go to the build dir
- all build tasks names should start with "build-"
- special website generation scripts do not belong to the core lib, but to a website repo/ website branch

timaschew:
Agree.
But the readme.md needs to be synchronized and the source of truth seems to be the master branch.
So either there should be a script which does a push from the master to the gh-pages which is located in the master branch
or there is a script which will pull the readme.md from the master branch to the gh-pages which is located in the gh-pages branch (only works after the commit of the readme file)

pithu:
yes, README.md of the master branch is the source for now, but the gh-pages update process keeps a manual process. And yes, in my opinion a webpage update script should go to the gh-pages branch, but i guess its not really helpful for now, because the .md files are slightly different.

Anyway, i think with your changes, we have a jekyll web page and we are well prepared for the future. Wen can add more content and pages without a problem and who knows, the gh-pages webpage might get its own content.

PR #29, opened 2016-03-20
Simplify deployment for gh-pages - part 2

timaschew:
This is part two of #28 

With this file structure Jekyll will compile all the html partials and markdown files.
It's triggered by each commit to gh-pages automatically.
You get an email if something breaks.

I removed the TOC and replaced it with auto-toc.js
The real content is included via a `include` command, usually it's easier for HTML snippets, but in this case we want to include a markdown, that's the reason to use `capture` and `markdownify`

This way the original file from the master branch can be used without any modifications.

With this setup you only need to sync two markdown files from the master branch.


timaschew:
Oh, here is a demo http://timaschew.github.io/js-joda/cheat-sheet.html

pithu:
:+1: sounds great, i will check it

pithu:
I only found two issues for now, internal links are still broken (same as before), and i miss my js-joda script at the cheat-sheet (easy to fix)

Great work @timaschew thanx a lot

PR #45, opened 2016-03-30
Updates to tests, cleanup, ...

phueper:


pithu:
Did you add the new classes to the public API at js-joda.js ? We should check the compressed size of the minified distribution afterwards. We might have to update the info in the README.

phueper:
No, i didn't run the dist build yet, we might also want a new version number? i'll check the size and want to update the readme anyway with the new domain classes

phueper:
after build-dist: 
```
ll dist                                                                                                                                                                                                                                                                                                    (master|✚3…)
total 2.1M
-rw-rw-r-- 1 phueper phueper 528K Mar 30 11:27 js-joda.js
-rw-rw-r-- 1 phueper phueper 1.3M Mar 30 11:27 js-joda.js.map
-rw-rw-r-- 1 phueper phueper 230K Mar 30 11:27 js-joda.min.js
-rw-rw-r-- 1 phueper phueper  36K Mar 26 22:22 js-joda.min.js.gz
```

pithu:
The gz file is a bit missleading. Its just a placeholder/ reminder for keeping an eye on compressed size.
 
I would recommend to remove the gz file from the repo. It is manually created/ updated anyway. 
We should add a npm run task for it, something like `npm run build-gz-check`. And then execute it manually from time to time and update the info in the README.

pithu:
btw: we reached 42.4 kB size now

phueper:
:+1:

PR #54, opened 2016-04-21
update `DateTimeFormater.ofPattern esdoc`

### LINKED ISSUE #53: Update DateTimeFormatterPattern documentation to indicate which conversion currently work
As mentioned in https://github.com/js-joda/js-joda/issues/34#issuecomment-212482961, the current documentation of the ofPattern method mentions all patterns but does not indicate which of the patterns do not currently work due to missing locale support

We should either remove these patterns or mark them to make it clear, that they are not currently supported

phueper:
closes #53


pithu:
nice trick to create a table format with esdoc

PR #69, opened 2016-07-08
Update dependencies

phueper:
Since esdoc 0.4.7 (and also eslint >=3) need node >= 4 i updated the travis build to node 4... 

this also fixes #66 but i think we need to think about wether we really want it ... 

on the one hand node 4.4.7 is the current  LTS , on the other hand it's kinda sad that we need to update node just because of the dev dependencies :/ 

Let's discuss this @pithu ?


pithu:
Is there a way to split the test runs? One just running the tests with an 0.12 node version, and another for all the build stuff with node 4.

phueper:
hmm... will check ...

phueper:
however, since we (dev) depend on eslint ... eslint >= 3 will probably not even `npm install` (since it depends on node >= 4)

pithu:
That is a nasty issue.

On the one hand there is no reason why js-joda should not run on an old node version. To guarantee that, automated tests are required. 

On the other hand we want to have the latest versions for developing. 

We have either prioritize whats more important or find a way to run tests with different node versions.

I cant imagine that we are the only one with that problem.

phueper:
Hmmm... What we could try would be to use docker... npm install and build in one container on node 4, npm test in another container on node 0.12 and share the node_modules volume between the two. That might work

phueper:
travis is very very cool :) ... i used the travis build matrix... now we build node 0.12, 4, 5, and latest (6)... only in latest are build-esdoc and test-coverage executed... i even experimented with an osx build, but that doesn't work yet... 

this might not solve the eslint > 3.0 problem... but let's solve that, when it comes :)

pithu:
Cool! 

Dont you think its enough to test against two node versions? Lets say node 0.10 and latest. Since we have no deps to the node api itself, we just have to ensure that our es5 code is running with an old v8 version.

phueper:
Ack... I changed it to build 0.10, 4 (LTS) and latest.

Issue #23, opened 2016-03-15
flow / TypeScript type annotations / bindings


phueper:
It would be nice to have flow and Typescript annotations / bindings to be able to type check the js-joda code itself but also give users the possibility to type check their usage of js-joda...
Even nicer would be if those could be generated from the esdoc comments.
Flow seems to have a way to have inline comments in e.g. function parameters but does not use the docblocks. I haven't checked for typescript yet

spencerwi:
Before seeing this issue, I started writing some to submit to DefinitelyTyped...but there's a _lot_ there and I'm not sure I have the pieces right.

I'll toss what I've got up in a gist so that it'll at least be a good start point in case y'all want to use it.

spencerwi:
Here's the gist: https://gist.github.com/spencerwi/03ac7f4dc1ab093166f0f676f8faa7e1

phueper:
Very cool contribution, thanks a lot! 
Would you mind creating a Pull Request for this? Then we could discuss in the PR if needed and also (credit where credit is due) your name would be on the commit as the author of this contribution. We would then get the file into the distribution so others can use it to typecheck. Also then we could maybe sync this to a commit to DefinitelyTyped so the Type Bindings could stay in sync?

jrnail23:
Along these lines, have y'all considered maybe incrementally rewriting js-joda in TypeScript?
It's a lot more of a natural conversion from Java than pure javascript, and in addition to the typechecking benefits, you'd have TypeScript definitions that would always be up to date.
If you have any interest at all, I'd be interested in pitching in for that effort.

The reason I'm bringing this up is that I'm finding myself depending on the existing typescript definitions to help me get some flowtype defs going for my own work, and I'm finding some things here and there that are either incomplete, or might not translate so well.

phueper:
Hi @jrnail23 

regarding typescript and/or flow, we would like to keep our codebase in "standard" ES6/7/... 
We do provide the typescript .d.ts files and would also provide flowtype definitions, however currently they are manually generated... i would really like to have a "single source of truth" for the interfaces (preferrably in our esdoc comments) and generate the typescript and flowtype definitions from that, but i haven't found an easy way to do that yet...

[...]

PR #81, opened 2016-08-16
Add WIP Typescript definitions (#23)

spencerwi:
Creating a PR with my work so far on the `.d.ts` definitions for this lib. There's a _lot_ to this lib, and while I went off the ESDoc pages, it's hard to know if I got things right right (there are some places where the signatures don't quite match between ESDoc and code, and some classes are not ESDoc'd). A lot of the not-really-user-facing classes are also not documented.

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -1,0 +1,886 @@
+declare abstract class TemporalAccessor {
+    get(field: TemporalField): number
+    query(query: TemporalQuery): any
+    range(field: TemporalField): ValueRange
+}
+declare abstract class Temporal extends TemporalAccessor {}
+
+declare abstract class Clock {
+    static fixed(fixedInstant: Instant, zoneOffset: ZoneOffset): Clock
+    static system(zone: ZoneId): Clock
+    static systemDefaultZone(): Clock
+    static systemUTC(): Clock
+
+    abstract instant(): Instant
+    abstract millis(): number
+    abstract zone(): any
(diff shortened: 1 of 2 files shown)


phueper:
Thanks a lot again @spencerwi !
I'm not much of an typescript expert ... but out of curiosity... is there any way we can use typescript to "verify" this file?
e.g. some kind of unit test or npm script that we could run that checks wether the definitions in this file match the class and function names that we define in the code?

I've seen that DefinitelyTyped also requires tests for the .d.ts files contributed (http://definitelytyped.org/guides/contributing.html) ... maybe something like that to verify the file?

Guria:
Yeah. You can install typescript as dependency and write some tests in ts files.

Guria:
There is no need to push typings in DefinetyleTyped if lib has own typings defined in package.json.

phueper:
FYI since i don't know how to add something to an existing (not mine) PR, i will probably create a new PR from my branch https://github.com/phueper/js-joda/tree/typescript_definitions including @spencerwi s commit and test files and other modifications 

as soon as a PR is coming out of that branch it will probably obsolete/close this PR

Guria:
Just keep it open. When you merge your pr, github will mark this one as merged too.

spencerwi:
So....anything more you need me to do here?

phueper:
No, currently not... unless you want to comment on the changes i did in my branch... any input is of course very much welcome

PR #83, opened 2016-08-18
Typescript definitions

phueper:
based on PR #81 provided by @spencerwi, 

restructured the typings file (namespace and module added) and added a test file as proposed by http://definitelytyped.org/guides/contributing.html

as the original PR states this is WIP but i think since it verifies with the samples from the CheatSheet it is a very good base to provide typescript typings. Maybe these could also be used to generate flow type definitions and to verify the esdoc comments in the future?

closes #81
 
This also relates to #23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_script:
   - sleep 3 # give xvfb some time to start
 script:
   - export
+  - npm run test-ts-definitions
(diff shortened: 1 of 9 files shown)


pithu:
👍

phueper:
:white_check_mark: done :)

PR #86, opened 2016-08-27
make typescript defenitions to be module definition

Guria:
Global typescript definition is intended for usage with DefinetelyTyped repo.
Definitions pointed with typings field in package.json must be in module format.

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -1465,6 +1465,4 @@ declare namespace JSJoda {
     function nativeJs(date: Date|any, zone?: ZoneId): TemporalAccessor;
 }
 
-declare module "JSJoda" {
-    export = JSJoda;
-}
+export = JSJoda;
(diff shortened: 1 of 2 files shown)


phueper:
Thx a lot... Just out of curiosity,  could you give me a pointer to where this is documented? I looked a little,  but only found the conventions as provided by DefinitelyTyped? Is this some kind of best practice?

Guria:
I am not sure I can find quickly where it is documented. But it is a way I know how node modules resolution work. We can call @blakeembrey to comment

phueper:
Hmkay, i like the import in your code much better without the reference line,  so  l'll pull this but would still be interested in documentation

Guria:
few links
https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Library%20Structures.md
https://github.com/Microsoft/TypeScript-Handbook/tree/master/pages/declaration%20files/templates

phueper:
:+1: thx

blakeembrey:
Also https://www.typescriptlang.org/docs/handbook/module-resolution.html and https://www.typescriptlang.org/docs/handbook/modules.html. If you're authoring a module, you shouldn't need to do `declare module` as that just creates a "global" definition - instead you can import/export at the top level for it to be "external" and used by TypeScript during resolution.

manland:
Thanks guys ;) A publish is scheduled ?

phueper:
i just published a new version including these changes

PR #88, opened 2016-08-30
Increase test coverage

phueper:
WIP: i want to get our coverage up some more... will add more tests here, but i want to let coveralls check this to see the improvement, thus i need this PR

The idea is to add reference (threetenbp ported) tests wherever more are needed/available, extend them if necessary and when the reference tests don't suffice, we create our own tests files and define tests there to get the coverage up

I discovered some minor edge case problems and also added some missing features e.g. in DateTimeBuilder (more fields now supported and tested)


phueper:
i'll pull this now to not get too far away from `master` ... more improvements coming up :)

PR #100, opened 2016-10-07
Provide a way to extend js-joda

pithu:


pithu:
Well, there is a test for the plugin code, but its excluded for karma, we can execute the test only with mocha.
Another side effect of the pr is that code coverage decreases, because the new plugTest.js requires the main entrypoint js-joda.js. We didn't test some files at all and there are in the stats now.

pithu:
I checked it, actually there was just one file added to the stats, i fixed it already yesterday.

phueper:
ah... i didn't realize that it's excluded for karma only but running in mocha...

pithu:
Oh my god your to fast ;)

I am still thinking about it. Is this the way we want to add plugins? 
The use function doesn't do a lot, since it is only exposing the public API you don't even need it.

On the other hand:
+ we provide a standard way to add plugins
+ chaijs is doing it the same way and its quite convenient
+ it prevents plugin code from being executed twice
+ we could expose more then the public API in the future
 
The idea is to add plugins the following way
```
    jsJoda = require('js-joda')
        .use(require('js-joda-timezone'))
        .use(require('js-joda-locale'))
        .use(require('js-joda-thaibuddhist'))
```

We could even think about splitting the js-joda code base into parts like
```
    jsJoda = require('js-joda-core')      // incl. LocalDate/ LocalDateTime/ Period/ etc.
        .use(require('js-joda-zoned'))     // incl. ZonedDateTime/ ZoneId/ etc
        .use(require('js-joda-additional'))  // incl. Year/ YearMonth/ etc.
        .use(require('js-joda-formater'))    // DateTimeFormatter/ etc.
```

phueper:
Sry...  I was in a merge mood 😂 i totally like it... It works,  it's tested,  let's work with it... If we want to further enhance it,  let's do it in master? No reason to leave it in a branch/PR i think?

pithu:
No problem, next time i will add a WorkInProgress label

phueper:
👍 

I like the splitting idea, possibly  reduces required size,  though I'm not sure how well it works... E.g. Some of the core classes do use formatter for parsing/ toString i think...

pithu:
I also believe that splitting isn't trivial

PR #107, opened 2016-11-02
Use Saucelabs to test browsers and parallelize travis build

phueper:
After #104  we decided to stop testing older node versions, the code we produce should really not behave any different on older node versions than on latest.
But it would be interesting to test the code in more (browser) environments.

So this PR

- drops the run with different node versions in travis, and replaces it with a single node version `--lts`, i.e. the latest stable LTS
- adds configuration for running the karma tests in saucelabs (other providers should work as well if we ever decide to move somewhere else from saucelabs
- the configuration currently needs SAUCE_USERNAME and SAUCE_ACCESS_KEY env entries to be set (when running local to test or when run in travis). For now that seems to be the easiest setup that works local, in forked repos and the main repo. Encrypted travis variables only work in the main repo so i decided not to use them for now
- to speed up the build the travis build matrix is used and we parallelize the build steps (actual build commands, node local tests, local karma tests (firefox, phantomks) and remote karma tests (chrome and ie9). This let's us have a result for the complete build in ~3-4 minutes instead of ~10-12 minutes when run sequentially  

Closes #104


karma.conf.js
```diff
@@ -1,24 +1,48 @@
...
+    
     config.set({
         files: [
-            {pattern: 'test/*Test.js'},
-            {pattern: 'test/**/*Test.js'},
-        ],
-        exclude: [
-            'test/plugTest.js'
+            {pattern: 'test/karmaWebpackTestEntry.js'}
         ],
```
pithu: I am waiting for karma webpack plugin, that allows you do configure file pattern and generates a single webpack bundle.

phueper: :+1:

Issue #20, opened 2016-03-01
Documentation is just copied from the Java world


pithu:
The documentation of js-joda is just copied from the https://github.com/ThreeTen/threetenbp java repo. The Threeten docu contains tons of very interesting and helpful information's not only about the lib itself but about date and time issues in general. But it contains references to java specific issue and java specific types, such as long/ int issues or thread-safe issues. And it contains javadoc tags, that are invalid for esdoc.

The only things we did so far, was to adapt the esdoc javascript parameters, so the esdoc docu and the IDE support for parameter looks good.

But the rest of the documentation needs to be reworked. The java issues must be removed/ adapted and javascript specific issues like floating-point, MAX/ MIN_SAFE_INTEGER must be added.

phueper:
actually some of the documentation formatting also does not seem to map very well to esdoc... e.g. the pattern table in DateTimeFormatterBuilder.appendPattern , and all {@code xxx} references

phueper:
I also just realized, that building the esdoc gives lots of warnings about signature mismatches :/

DanielSWolf:
The warnings about signature mismatches seem to have been fixed by pull request #87. However, all `@code` and `@link` tags still get copied verbatim.

DanielSWolf:
Here's what I found:

* `@link` is working fine, it's just sometimes used incorrectly:
  * It must not contain extra text. For instance, `{@link ZoneRules rules}` should be `{@link ZoneRules} rules`.
  * It must contain full method names. Instead of `{@link #withEarlierOffsetAtOverlap()}`, use `{@link ZonedDateTime#withEarlierOffsetAtOverlap}`. No parentheses, `#` for instance methods, `.` for static methods.
* `@code` is not supported by ESDoc. It can be replaced by the following:
  * It seems like most of the time, `@code` is currently used to refer to a class, so `@link` should be used instead.
  * In those places where `monospaced` formatting is desired, Markdown (backticks) can be used.

Using search&replace and a little time, it should be easy to get good-looking output.

PR #111, opened 2016-11-08
Documentation fixes

DanielSWolf:
This PR fixes various technical issues with the documentation, as discussed in #20. There are still some issues left regarding `@link`s to non-existent properties, but for the most part, the generated documentation is looking nice and clean now.

As an added bonus, I've converted all the ugly HTML tags in the documentation comments to Markdown.

Note that this PR does *not* correct Java-related content in the documentation, so it's not a complete fix for #20.


phueper:
thx a lot @DanielSWolf  ... great work :+1:

PR #113, opened 2016-11-14
replace var with const/let

chauthai:


src/DayOfWeek.js
```diff
@@ -71,12 +71,14 @@ export class DayOfWeek extends Temporal {
      * @returns {DayOfWeek}
      */
     static valueOf(name) {
-        for(var ordinal=0; ordinal < ENUMS.length; ordinal++){
+        let ordinalResult = 0;
```
phueper: i think the additional `ordinalResult` variable is not necessary, imho it would be nicer to initialize the ordinal variable outside of the loop, modify it in the loop and then return it after the loop

src/LocalTime.js
```diff
@@ -427,6 +427,7 @@ export class LocalTime extends Temporal /** implements Temporal, TemporalAdjuste
      * @private
      */
     _get0(field) {
+        const ham = MathUtil.intMod(this._hour, 12);
```
phueper: this calculation is unnecessary for all cases except CLOCK_HOUR_OF_AMPM, please only do the calculation inside the corresponding case

src/temporal/IsoFields.js
```diff
@@ -777,9 +777,9 @@ class Unit extends TemporalUnit {
      * @returns {number}
      */
     addTo(temporal, periodToAdd) {
+        const added = MathUtil.safeAdd(temporal.get(WEEK_BASED_YEAR), periodToAdd);
```
phueper: this calculation is only necessary for WEEK_BASED_YEARS, please only do the calculation inside the case

src/Duration.js
```diff
@@ -353,7 +353,7 @@ export class Duration extends TemporalAmount /*implements TemporalAmount, Compar
             if (parsed[0] === '+') {
                 parsed = parsed.substring(1);
             }
-            var val = parseFloat(parsed);
+            const val = parseFloat(parsed);
```
phueper: this could be inlined i think, the variable has no benefit?

src/Instant.js
```diff
@@ -655,10 +655,10 @@ export class Instant extends Temporal {
         if ((secondsToAdd | nanosToAdd) === 0) {
             return this;
         }
-        var epochSec = this._seconds + secondsToAdd;
+        let epochSec = this._seconds + secondsToAdd;
         epochSec = epochSec + MathUtil.intDiv(nanosToAdd, LocalTime.NANOS_PER_SECOND);
-        var _nanosToAdd = nanosToAdd % LocalTime.NANOS_PER_SECOND;
-        var nanoAdjustment = this._nanos + _nanosToAdd;
+        const _nanosToAdd = nanosToAdd % LocalTime.NANOS_PER_SECOND;
+        const nanoAdjustment = this._nanos + _nanosToAdd;
```
phueper: could be inlined

src/LocalDate.js
```diff
@@ -1473,12 +1473,12 @@ export class LocalDate extends ChronoLocalDate{
...
     atStartOfDayWithZone(zone) {
         requireNonNull(zone, 'zone');
-        var ldt = this.atTime(LocalTime.MIDNIGHT);
+        let ldt = this.atTime(LocalTime.MIDNIGHT);
         // need to handle case where there is a gap from 11:30 to 00:30
         // standard ZDT factory would result in 01:00 rather than 00:30
         if (zone instanceof ZoneOffset === false) {
-            var rules = zone.rules();
-            var trans = rules.transition(ldt);
+            const rules = zone.rules();
```
phueper: could be inlined

src/LocalDate.js
```diff
@@ -1496,9 +1496,9 @@ export class LocalDate extends ChronoLocalDate{
      * @return {number} the Epoch Day equivalent to this date
      */
     toEpochDay() {
-        var y = this.year();
-        var m = this.monthValue();
-        var total = 0;
+        const y = this.year();
```
phueper: could be replaced with access to the "private" field instead of the function call?

src/LocalDate.js
```diff
@@ -1496,9 +1496,9 @@ export class LocalDate extends ChronoLocalDate{
...
+        const m = this.monthValue();
```
phueper: could be replaced with access to the "private" field instead of the function call?

src/LocalDate.js
```diff
@@ -1671,13 +1671,13 @@ export class LocalDate extends ChronoLocalDate{
      * @return {string} a string representation of this date, not null
      */
     toString() {
-        var dayString, monthString, yearString;
+        let dayString, monthString, yearString;
 
-        var yearValue = this.year();
-        var monthValue = this.monthValue();
-        var dayValue = this.dayOfMonth();
+        const yearValue = this.year();
```
phueper: could be replaced with access to the "private" field instead of the function call?

src/LocalDate.js
```diff
@@ -1671,13 +1671,13 @@ export class LocalDate extends ChronoLocalDate{
...
+        const monthValue = this.monthValue();
```
phueper: could be replaced with access to the "private" field instead of the function call?

src/LocalDate.js
```diff
@@ -1671,13 +1671,13 @@ export class LocalDate extends ChronoLocalDate{
...
+        const dayValue = this.dayOfMonth();
```
phueper: could be replaced with access to the "private" field instead of the function call?

src/LocalDateTime.js
```diff
@@ -210,8 +210,8 @@ implements Temporal, TemporalAdjuster, Serializable */ {
         requireNonNull(instant, 'instant');
         requireInstance(instant, Instant, 'instant');
         requireNonNull(zone, 'zone');
-        var rules = zone.rules();
-        var offset = rules.offset(instant);
+        const rules = zone.rules();
```
phueper: could be inlined?

src/format/DateTimeFormatter.js
```diff
@@ -248,7 +248,7 @@ export class DateTimeFormatter {
      * @throws IllegalArgumentException if the pattern is invalid
      * @see DateTimeFormatterBuilder#appendPattern(String)
      * @example
-     * var s = LocalDate.parse('2016-04-01').format(DateTimeFormatter.ofPattern('d MM yyyy'));
+     * const s = LocalDate.parse('2016-04-01').format(DateTimeFormatter.ofPattern('d MM yyyy'));
```
phueper: this example should not use ES6 notation imho, but keep ES5 var notation

phueper:
Hi @chauthai ... thanks  a lot again for the changes!

PR #120, opened 2016-11-30
parse dayOfMonth and month to int to validate dayOfMonth correctly

axelwathne:
LocalDate accepts strings as parameters, but circumvents validation because the parameters are simply compared to numbers, which leads to '32' > 31 = false.

I added a test to the non-reference tests, because this is really a type safety issue, which is not an issue in threeten.

### DIFF
--- a/src/LocalDate.js
+++ b/src/LocalDate.js
@@ -303,15 +303,10 @@ export class LocalDate extends ChronoLocalDate{
         ChronoField.MONTH_OF_YEAR.checkValidValue(month);
         ChronoField.DAY_OF_MONTH.checkValidValue(dayOfMonth);
         
-        if(typeof month === 'string'){
-            month = parseInt(month);
-        }
-        
-        if(typeof dayOfMonth === 'string'){
-            dayOfMonth = parseInt(dayOfMonth);
-        }
-        
         if (dayOfMonth > 28) {
+            if(typeof month === 'string'){
+                month = parseInt(month);
+            }
             dom = 31;
             switch (month) {
                 case 2:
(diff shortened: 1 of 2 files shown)


phueper:
Hi @axelwathne, thanks for reporting this

According to http://es5.github.io/#x11.8 i think that relational comparisons between strings (that parse to numbers) and numbers should work, see https://runkit.com/phueper/583f53ecb3ef0f0013baa687 

However in this case the switch statement is the problem i think, so the fix is correct.

However, i would like to see more tests for this case, e.g. a separate 'it' with successfully parsed numbers compared to an LocalDate from actual numbers, and also some more error cases (strings that cannot be parsed successfully). Would you mind adding these tests, maybe taking some hints from the reference tests?

Regards, Pattrick

pithu:
What about moving the type conversion into the constructor. We doing it there anyway but after the validation. To fully support string values we should replace `MathUtil.safeZero` by `MathUtil.parseInt`.
could be something like: 
```
constuctor ...
        ...
        this._year = MathUtil.parseInt(year);
        this._month = MathUtil.parseInt(month);
        this._day = MathUtil.parseInt(dayOfMonth);
        LocalDate._validate(this._year, this._month, this._dayOfMonth);
```
And yes, an extra test case for string support would be great.

axelwathne:
Offcourse, you are right, I got a little ahead of myself and thought I had actually instantiated a date with day 32. The conversion for dayOfMonth is not needed, so we can omit the type checks altogether for dayOfMonth =< 28.

I'll add some tests too, but checking the entire library for this kind of type error would be a bit too much for me right now.

axelwathne:
@pithu I agree, the type conversion will be done eventually anyway, so all methods that take strings that eventually will be numbers should convert immediately. That way, you remove the possibility of errors like this by modifiying the input to code ported from a type safe language instead of altering the code in a lot of places. And one explicit parseInt should be faster than a bunch of implicit ones.

phueper:
i like @pithu s proposal, but let's do that in a separate PR, will merge this one

PR #126, opened 2017-01-19
Fix LocalTime.parse

jamiewinder:
Seconds argument of `LocalTime.parse` is a `DateTimeFormatter`, not a string.


phueper:
:+1:  correct, for consistency, could you please also correct this in https://github.com/js-joda/js-joda/blob/master/src/LocalTime.js#L257 ?

jamiewinder:
Done. I've followed the convention from LocalDate.parse. I think it's accurate..?

phueper:
thx!

PR #159, opened 2017-05-12
Added `boolean` return values to type definitions

paplorinc:
* Added `boolean` return values to `equals`/`isEqual`, `isBefore`/`isAfter`
* Added `string` type to `toString`
* Formatted union types uniformly

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -50,7 +50,7 @@ declare namespace JSJoda {
 
         getLong(field: TemporalField): number
 
-        isSupported(field: TemporalField): number
+        isSupported(field: TemporalField): boolean
 
         minus(days: number): DayOfWeek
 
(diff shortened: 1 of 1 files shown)


pithu:
Hello @paplorinc 

thx for contributing again. Did you check the corresponding type definitions in the esdoc and fix it there as well ?

Its because we want the esdoc to be the leading source for type definitions. In an ideal world we would generate  typescript ( and flow, ...) definitions from the esdoc

paplorinc:
Thanks @pithu, could you help me out with how I should go about doing that? :)

pithu:
eg. For your fix https://github.com/js-joda/js-joda/pull/159/files#diff-e243a30e5201f226a48a91f2961294fbR53 find the corresponding esdoc type definition for DayOfWeek.isSupported  here https://github.com/js-joda/js-joda/blob/master/src/DayOfWeek.js#L180. In that case the esdoc is already correct.

paplorinc:
@pithu, thanks, checked all `esdoc` definitions and they were correct (though the type wasn't explicitly stated in all cases, just "`true` if this `false` if that")

Issue #162, opened 2017-05-15
Add static `min` and `max` to `LocalDate/Time`


paplorinc:
```javascript
function min<T extends {isBefore: (_: T) => boolean}>(first: T, second: T): T {
  return first.isBefore(second) ? first : second;
}
```
```javascript
function max<T extends {isAfter: (_: T) => boolean}>(first: T, second: T) {
  return first.isAfter(second) ? first : second;
}
```

Issue #161, opened 2017-05-15
Add `isAfterOrEqual` and `isBeforeOrEqual` to `LocalDate/Time`


paplorinc:
`isAfterOrEqual` = `!this.isBefore(other)`
`isBeforeOrEqual` = `!this.isAfter(other)`

Similarly to https://github.com/js-joda/js-joda/blob/4a74d6e61f2f3dde70aad4c5cf97f9e38a1e437c/test/plugTest_mochaOnly.js#L17

PR #164, opened 2017-05-15
Added min/max and ...OrEqual to LocalDate

paplorinc:
Request for comment for https://github.com/js-joda/js-joda/issues/161 and https://github.com/js-joda/js-joda/issues/162.
If you think this is a viable direction, I can add it to `LocalTime` and `ZonedDateTime` also, and generate the final deliverable.

### DIFF
--- a/CheatSheet.md
+++ b/CheatSheet.md
@@ -453,12 +453,17 @@ t.truncatedTo(ChronoUnit.DAYS);      // '00:00'
 ### Compare LocalTime instances
 
 ```javascript
+var today = LocalDate.now(), tomorrow = today.plusDays(1);
+LocalDate.min(today, tomorrow); // today
+LocalDate.max(today, tomorrow); // tomorrow
 
 var t1 = LocalTime.parse('11:55:42')
 var t2 = t1.plusHours(2);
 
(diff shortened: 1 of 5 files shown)


phueper:
Hi @paplorinc  ... thanks for the PR...

we want js-joda to be a port of the joda/JSR-310 API, specifically we chose the threetenbp project to port to JavaScript. In this repo (i.e. the "base" js-joda repo) we are pretty strict to follow the API of our upstream project and will not accept any changes/enhancements to that API.
So, i am sorry, but we will not accept this PR here. 

I can see the advantage of the changes you propose though. We tried to create a plugin system to enhance the base js-joda API and functionality. So i think that either you could create your own plugin to js-joda and add the functionality there. 
Or we could think about adding your enhancements to the  https://github.com/js-joda/js-joda-extra plugin  repository. I think it would fit very nicely in the js-joda-extra repo, we would have to think about how we can enhance the `.d.ts` file from a plugin to add the definitions, but i'd hope that should be possible.

Regards, Pattrick

paplorinc:
Hey Pattrick, it's ok, feel free to add it to `extra` eventually.
Or if it matures, I may add it myself later :)

PR #179, opened 2017-08-12
typescript definition: Fix addTo/subtractFrom

benny-medflyt:
This changes the `addTo` and `subtractTo` functions, so that whatever type of Temporal object you pass in, you get the same type returned.

This is needed so that when you write something like:

```typescript
var in_a_minute = Duration.ofSeconds(60).addTo(Instant.now());
```

then the variable `in_a_minute` is of type "Instant", with all of the Instant goodness available, rather than the current behavior where `in_a_minute` ends up as the (useless) type "Temporal".

I believe that there are other usages of Temporal where the types can be tightened up, but this pull request is at least a start. (another candidate for improvement are the `between` functions.. for example, is the following sensible? or must both arguments be of the same type of Temporal?)

```typescript
Duration.between(Instant.now(), LocalTime.now())
```

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -65,13 +65,13 @@ declare namespace JSJoda {
         value(): number
     }
     class TemporalAmount {
-        addTo(temporal: Temporal): Temporal
+        addTo<T extends Temporal>(temporal: T): T
 
         get(unit: TemporalUnit): number
 
         units(): TemporalUnit[]
 
-        subtractFrom(temporal: Temporal): Temporal
+        subtractFrom<T extends Temporal>(temporal: T): T
     }
     class Duration extends TemporalAmount {
         static ZERO: Duration
@@ -98,7 +98,7 @@ declare namespace JSJoda {
 
         abs(): Duration
 
-        addTo(temporal: Temporal): Temporal
+        addTo<T extends Temporal>(temporal: T): T
 
         compareTo(otherDuration: Duration): number
 
@@ -158,7 +158,7 @@ declare namespace JSJoda {
 
         seconds(): number
 
-        subtractFrom(temporal: Temporal): Temporal
+        subtractFrom<T extends Temporal>(temporal: T): T
 
         toDays(): number
 
@@ -646,7 +646,7 @@ declare namespace JSJoda {
         toString(): string
     }
     class TemporalUnit {
-        addTo(temporal: Temporal, amount: number): Temporal
+        addTo<T extends Temporal>(temporal: T, amount: number): T
 
         between(temporal1: Temporal, temporal2: Temporal): number
 
@@ -677,7 +677,7 @@ declare namespace JSJoda {
         static ERAS: ChronoUnit
         static FOREVER: ChronoUnit
 
-        addTo(temporal: Temporal, amount: number): Temporal
+        addTo<T extends Temporal>(temporal: T, amount: number): T
 
         between(temporal1: Temporal, temporal2: Temporal): number
 
@@ -1002,7 +1002,7 @@ declare namespace JSJoda {
 
         static parse(text: string): Period
 
-        addTo(temporal: Temporal): Temporal
+        addTo<T extends Temporal>(temporal: T): T
 
         chronology(): IsoChronology
 
@@ -1042,7 +1042,7 @@ declare namespace JSJoda {
 
         plusYears(yearsToAdd: number): Period
 
-        subtractFrom(temporal: Temporal): Temporal
+        subtractFrom<T extends Temporal>(temporal: T): T
 
         toJSON(): string
 


phueper:
Hi @benny-medflyt 

Thanks a lot for your Pull Request, this looks nice and makes sense i guess.

Would you mind looking for and improving the other places you mentioned? i think your example of `between` will work for any objects as long as they extend `Temporal` 

Regards, Pattrick

benny-medflyt:
I wrote a script to check all the combinations. Here are the results for `Duration.addTo` and `Duration.between`:

```
Duration.addTo ERROR LocalDate
Duration.addTo OK:   LocalTime (LocalTime)
Duration.addTo OK:   LocalDateTime (LocalDateTime)
Duration.addTo OK:   ZonedDateTime (ZonedDateTime)
Duration.addTo OK:   Instant (Instant)
Duration.between ERROR LocalDate LocalDate
Duration.between ERROR LocalDate LocalTime
Duration.between ERROR LocalDate LocalDateTime
Duration.between ERROR LocalDate ZonedDateTime
Duration.between ERROR LocalDate Instant
Duration.between ERROR LocalTime LocalDate
Duration.between OK:   LocalTime LocalTime (Duration)
Duration.between OK:   LocalTime LocalDateTime (Duration)
Duration.between OK:   LocalTime ZonedDateTime (Duration)
Duration.between ERROR LocalTime Instant
Duration.between ERROR LocalDateTime LocalDate
Duration.between ERROR LocalDateTime LocalTime
Duration.between OK:   LocalDateTime LocalDateTime (Duration)
Duration.between OK:   LocalDateTime ZonedDateTime (Duration)
Duration.between ERROR LocalDateTime Instant
Duration.between ERROR ZonedDateTime LocalDate
Duration.between ERROR ZonedDateTime LocalTime
Duration.between ERROR ZonedDateTime LocalDateTime
Duration.between OK:   ZonedDateTime ZonedDateTime (Duration)
Duration.between ERROR ZonedDateTime Instant
Duration.between ERROR Instant LocalDate
Duration.between ERROR Instant LocalTime
Duration.between ERROR Instant LocalDateTime
Duration.between OK:   Instant ZonedDateTime (Duration)
Duration.between OK:   Instant Instant (Duration)
```

## Duration.addTo

  - I was correct that `Duration.addTo` always returns the same type that it got in.
  - However, notice that calling `Duration.addTo` with a `LocalDate` is illegal. The current TypeScript definition allows this. It's not clear to me how to fix this with the current inheritance hierarchy.
    - The easiest is to give in the function definition an explicit list of the types allowed (`LocalTime`, `LocalDateTime`, `ZonedDateTime`, `Instant`). This would work, but I'm not sure if js-joda library is meant to be extensible in terms of allowing users of the library to extend or create their own types. If so, then wouldn't be able to use the `addTo` function on their own types (without a cast). If such extensibility is not intended (and I personally don't see any virtue in it) then again, there is an easy fix

## Duration.between

  - Here the situation is more complicated. It appears that mixing and matching some combinations of types is indeed allowed.
  - Not that here too, `LocalDate` is not allowed at all (and typescript let's it slip through..)
  - There are worrying results. Some pairs are only allowed in certain orders. For example, this works:

        Duration.between(Instant.now(), ZonedDateTime.now())

    while this throws an error:

        Duration.between(ZonedDateTime.now(), Instant.now())

    And there are a few more similar examples as you can see in the results. My guess is that these are all actually bugs in js-joda, and that the only valid intended use of `Duration.between` is when both arguments are of the same type.

# Script:

```javascript
var LocalDate = require("js-joda").LocalDate;
var LocalTime = require("js-joda").LocalTime;
var LocalDateTime = require("js-joda").LocalDateTime;
var ZonedDateTime = require("js-joda").ZonedDateTime;
var Instant = require("js-joda").Instant;
var Duration = require("js-joda").Duration;

function tryDurationAddTo(a) {
    try {
        var r = Duration.ofDays(1).addTo(a);
        console.log("Duration.addTo OK:  ", a.constructor.name, "(" + r.constructor.name + ")");
    } catch (e) {
        console.log("Duration.addTo ERROR", a.constructor.name);
    }
}

function tryDurationBetween(a, b) {
    try {
        var r = Duration.between(a, b);
        console.log("Duration.between OK:  ", a.constructor.name, b.constructor.name, "(" + r.constructor.name + ")");
    } catch (e) {
        console.log("Duration.between ERROR", a.constructor.name, b.constructor.name);
    }
}

tryDurationAddTo(LocalDate.now());
tryDurationAddTo(LocalTime.now());
tryDurationAddTo(LocalDateTime.now());
tryDurationAddTo(ZonedDateTime.now());
tryDurationAddTo(Instant.now());

tryDurationBetween(LocalDate.now(), LocalDate.now());
tryDurationBetween(LocalDate.now(), LocalTime.now());
tryDurationBetween(LocalDate.now(), LocalDateTime.now());
tryDurationBetween(LocalDate.now(), ZonedDateTime.now());
tryDurationBetween(LocalDate.now(), Instant.now());

tryDurationBetween(LocalTime.now(), LocalDate.now());
tryDurationBetween(LocalTime.now(), LocalTime.now());
tryDurationBetween(LocalTime.now(), LocalDateTime.now());
tryDurationBetween(LocalTime.now(), ZonedDateTime.now());
tryDurationBetween(LocalTime.now(), Instant.now());

tryDurationBetween(LocalDateTime.now(), LocalDate.now());
tryDurationBetween(LocalDateTime.now(), LocalTime.now());
tryDurationBetween(LocalDateTime.now(), LocalDateTime.now());
tryDurationBetween(LocalDateTime.now(), ZonedDateTime.now());
tryDurationBetween(LocalDateTime.now(), Instant.now());

tryDurationBetween(ZonedDateTime.now(), LocalDate.now());
tryDurationBetween(ZonedDateTime.now(), LocalTime.now());
tryDurationBetween(ZonedDateTime.now(), LocalDateTime.now());
tryDurationBetween(ZonedDateTime.now(), ZonedDateTime.now());
tryDurationBetween(ZonedDateTime.now(), Instant.now());

tryDurationBetween(Instant.now(), LocalDate.now());
tryDurationBetween(Instant.now(), LocalTime.now());
tryDurationBetween(Instant.now(), LocalDateTime.now());
tryDurationBetween(Instant.now(), ZonedDateTime.now());
tryDurationBetween(Instant.now(), Instant.now());
```

pithu:
Hi, 

nice Pull Request! 

It should make  sense for the methods `addTo` and `subtractFrom` of domains `Duration` and `Period`. Interesting that neither `threeten` nor the `jdk8` uses the generic definition here. But i cant see any problems with it.

@phueper well we said that we want have esdoc as the leading source for type definitions. Since esdoc has nor support for generics i think we should leave that track.

With `between`  things get a bit more complicated. I guess its not a bug and i guess that the jdk behaves exactly the same for your above example. In short, it's because you can't compare apple with pears. And unfortunately i don't see anyway you can express this with generics type definitions. 

Threeten distinguish between time-based duration (domain `Duration`) and date-based duration (domain `Period`). That's the reason why `Duration.addTo` with a `LocalDate` fails, you have to use `Period.addTo` in that case.

`Duration.between` uses `Temporal.until` internally. It calls `firstArgument.until(secondArgument)`. And until tries to extract/convert the second argument into the same type as the first argument.  For example `LoacalTime.until(temporal)` tries to extract the localTime `from` the temporal. That works for any domain that has a `LocalTime` like a`LocalDateTime` but the other way it fails because you cant extract the `LocalDateTime` of a `LocalTime`.

Issue #150, opened 2017-04-05
some classes not listed in dist


henryw374:
Hi,

Navigating under the JSJoda object I get when importing  _https://rawgit.com/js-joda/js-joda/master/dist/js-joda.js_ , I see it is missing some things from the api, for example **Temporal** or **TemporalAccessor**. Would it be possible to have a way to access these things other than via this sort of thing?

`JSJoda.DayOfWeek.FRIDAY.__proto__`

Thanks

phueper:
Hi @henryw374,

thanks for reporting... the "scope" of the js-joda.js is for users of the library... for our "plugins" e.g. js-joda-locale, we are actually thinking about how we can give access to more parts of the API that "normal" users don't necessarily require ... Could you give us an example to understand how you would use the `Temporal` and `TemporalAccessor` classes... it might help us decide how to handle these :)

Regards, Pattrick

henryw374:
Ok cool so this would be useful in creating a Clojurescript wrapper for js-joda. Currently there is a  [Clojure (jvm) wrapper for this](https://github.com/dm3/clojure.java-time) and the plan is to provide the same api for a Clojurescript version. 

This new wrapper will ideally share much of the code of the Clojure version and that version already uses things like Temporal for polymorphic dispatch.

PR #183, opened 2017-09-05
Changes needed for js-joda-locale implementation

phueper:
These changes are needed for js-joda-locale implementation, e.g. to be able to override the unimplemented functions in the plugin, also export some needed classes

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 dist/*
 build/*
+tmp/*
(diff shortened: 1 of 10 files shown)


pithu:
One thing after first sight: What about exporting some more public interfaces as requested in issue #150 like `Temporal`, `TemporalAmount`, `TemporalUnit`

phueper:
Ah... good point... will check it

pithu:
Should we even export all classes interfaces that are public in threeten?

phueper:
yes, i think that makes sense... i will do another commit on this PR, exporting everything that is public in threeten and even more internal classes that *migh* be used by plugins

PR #188, opened 2017-10-09
Add missing TypeScript definitions for YearMonth class

lpatalas:
This pull request:

* adds missing method definitions for `YearMonth` class
* splits `plus` and `minus` definitions into separate overloads
* adds tests for `YearMonth` in `js-joda-tests.ts`

Best regards,
Łukasz

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -1163,18 +1163,73 @@ declare namespace JSJoda {
         static from(temporal: TemporalAccessor): YearMonth
 
         static now(zoneIdOrClock?: ZoneId|Clock): YearMonth
+        static now0(): YearMonth
+        static nowZoneId(zone: ZoneId): YearMonth
+        static nowClock(clock: Clock): YearMonth
 
         static of(year: number, monthOrNumber: Month|number): YearMonth
+        static ofNumberMonth(year: number, month: Month): YearMonth
+        static ofNumberNumber(year: number, month: number): YearMonth
 
         static parse(text: string, formatter?: DateTimeFormatter): YearMonth
+        static parseString(text: string): YearMonth
+        static parseStringFormatter(text: string, formatter: DateTimeFormatter): YearMonth
 
-        minus(amountOrNumber: TemporalAmount|number, unit?: TemporalUnit): YearMonth
+        constructor(year: number, month: number)
 
-        plus(amountOrNumber: TemporalAmount|number, unit?: TemporalUnit): YearMonth
+        minus(amount: TemporalAmount): YearMonth
+        minus(amountToSubtract: number, unit: TemporalUnit): YearMonth
+        minusAmount(amount: TemporalAmount): YearMonth
+        minusAmountUnit(amountToSubtract: number, unit: TemporalUnit): YearMonth
+        minusYears(yearsToSubtract: number): YearMonth
+        minusMonths(monthsToSubtract: number): YearMonth
 
-        with(adjusterOrField: TemporalAdjuster|TemporalField, value?: number): YearMonth
+        plus(amount: TemporalAmount): YearMonth
+        plus(amountToAdd: number, unit: TemporalUnit): YearMonth
+        plusAmount(amount: TemporalAmount): YearMonth
+        plusAmountUnit(amountToAdd: number, unit: TemporalUnit): YearMonth
+        plusYears(yearsToAdd: number): YearMonth
+        plusMonths(monthsToAdd: number): YearMonth
+
+        with(adjuster: TemporalAdjuster): YearMonth
+        with(field: TemporalField, value: number): YearMonth
+        withYearMonth(newYear: number, newMonth: number): YearMonth
+        withAdjuster(adjuster: TemporalAdjuster): YearMonth
+        withFieldValue(field: TemporalField, newValue: number): YearMonth
+        withYear(year: number): YearMonth
+        withMonth(month: number): YearMonth
 
         isSupported(fieldOrUnit: TemporalField|ChronoUnit): boolean
+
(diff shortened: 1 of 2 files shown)


dist/js-joda.d.ts
```diff
@@ -1163,18 +1163,73 @@ declare namespace JSJoda {
         static from(temporal: TemporalAccessor): YearMonth
 
         static now(zoneIdOrClock?: ZoneId|Clock): YearMonth
+        static now0(): YearMonth
+        static nowZoneId(zone: ZoneId): YearMonth
+        static nowClock(clock: Clock): YearMonth
```
phueper: these methods are our idea of "overloading" the `now` method. Since these are not part of the threeten/joda API,  we do not consider these functions as public in the API and i think they should thus not be defined in the typescript definitions.
Since this is not obvious to a user of js-joda not familiar with the threeten API, i think we will have to *also* mark these methods as `private` in the esdoc comments to make it more obvious.
And... to make it clear, you should *not* use these methods, but rather use the `now` method which will (depending on argument count and/or argument type delegate the call to the appropriate overloaded method. And i think the typescript definition of the `now` method allows you to pass any of the arguments of the overloaded methods (none for `now0`, ZoneId or Clock for `nowZoneId` and `nowClock`

dist/js-joda.d.ts
```diff
@@ -1163,18 +1163,73 @@ declare namespace JSJoda {
...
 
         static of(year: number, monthOrNumber: Month|number): YearMonth
+        static ofNumberMonth(year: number, month: Month): YearMonth
+        static ofNumberNumber(year: number, month: number): YearMonth
```
phueper: see first comment on `now`/`now0`/`nowZoneId`/`nowClock`

dist/js-joda.d.ts
```diff
@@ -1163,18 +1163,73 @@ declare namespace JSJoda {
...
 
         static parse(text: string, formatter?: DateTimeFormatter): YearMonth
+        static parseString(text: string): YearMonth
+        static parseStringFormatter(text: string, formatter: DateTimeFormatter): YearMonth
```
phueper: see first comment on `now`/`now0`/`nowZoneId`/`nowClock`

dist/js-joda.d.ts
```diff
@@ -1163,18 +1163,73 @@ declare namespace JSJoda {
...
+        plusYears(yearsToAdd: number): YearMonth
+        plusMonths(monthsToAdd: number): YearMonth
+
+        with(adjuster: TemporalAdjuster): YearMonth
+        with(field: TemporalField, value: number): YearMonth
+        withYearMonth(newYear: number, newMonth: number): YearMonth
+        withAdjuster(adjuster: TemporalAdjuster): YearMonth
+        withFieldValue(field: TemporalField, newValue: number): YearMonth
+        withYear(year: number): YearMonth
+        withMonth(month: number): YearMonth
```
phueper: see first comment on `now`/`now0`/`nowZoneId`/`nowClock`

test/typescript_defintions/js-joda-tests.ts
```diff
@@ -442,4 +445,84 @@ function test_Duration() {
...
+function test_YearMonth() {
+    YearMonth.from(YearMonth.now());
+    YearMonth.from(LocalDate.now());
+    
+    YearMonth.now();
+    YearMonth.now(ZoneId.systemDefault());
+    YearMonth.now(Clock.systemUTC());
+    YearMonth.now0();
+    YearMonth.nowZoneId(ZoneId.systemDefault());
+    YearMonth.nowClock(Clock.systemUTC());
```
phueper: see first comment on `now`/`now0`/`nowZoneId`/`nowClock`

dist/js-joda.d.ts
```diff
@@ -1163,18 +1163,73 @@ declare namespace JSJoda {
...
+        static ofNumberNumber(year: number, month: number): YearMonth
 
         static parse(text: string, formatter?: DateTimeFormatter): YearMonth
+        static parseString(text: string): YearMonth
+        static parseStringFormatter(text: string, formatter: DateTimeFormatter): YearMonth
+
+        constructor(year: number, month: number)
+
+        minus(amount: TemporalAmount): YearMonth
+        minus(amountToSubtract: number, unit: TemporalUnit): YearMonth
```
phueper: i am not a typescript user, but what would be the advantage of these split definitions, instead of the optional/overloaded definitions that we had in the file originally? To me it seems more clear with the overloaded definition?

lpatalas: The JavaScript API as I see it is basically:

* You must specify one or two parameters
* If you specify one parameter then it's treated as `TemporalAmount`
* If you specify two parameters first one is `number` and the second one is `TemporalUnit`

With your original definition you could make this two mistakes:

* Pass `number` and don't specify unit, because second argument is always optional
* Pass `TemporalAmount` and `TemporalUnit` as second parameter and the method would assume that first one is `number`

After the change it will not compile and only the two correct combinations of arguments are allowed.

dist/js-joda.d.ts
lpatalas: I wasn't aware of this. I will make the necessary changes later today.

dist/js-joda.d.ts
phueper: True... good point. But that would mean we should change the other optional/overloaded definitions in the typescript file as well ... 

while our overloading *should* check this and cause an error at runtime, your change would allow the error to already occur at compile time... i think i like that :)

lpatalas:
I removed the private methods from `YearMonth` definition. It should be ok now.

PR #202, opened 2017-12-04
Make `TemporalAdjuster` an interface (not a class)

ianp:
Further work for js-joda#23.

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -25,7 +25,7 @@ declare namespace JSJoda {
 
         abstract zone(): any
     }
-    class DayOfWeek extends Temporal {
+    class DayOfWeek extends Temporal implements TemporalAdjuster {
         static MONDAY: DayOfWeek
         static TUESDAY: DayOfWeek
         static WEDNESDAY: DayOfWeek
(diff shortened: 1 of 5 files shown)


ianp:
Hmm, that's odd, the tests passed when I ran them locally. I'll take a look at it over the weekend and get back to you.

ianp:
OK, still not ready to merge, but I think that I see the problem now and want to run this solution by you before pressing ahead.

I've added `plus` and `minus` methods to `Temporal` so that the tests can be rewritten to just use the methods on this abstract class, and both `Month` and `DayOfWeek` have been modified to extend `TemporalAccessor` rather than `Temporal`.

Obviously a complete solution would need the corresponding `minus` methods, as well as all of the jsdoc comments, but does this seem OK to you?

phueper:
Hmmm... i'm not quite sure about this, i would like to get @pithu s opiniion as well... 

but i think it goes in the right direction... we want to stay close to our upstream threeten... in threeten Temporal is an interface, which we can't define in plain JS... but making all Temporal methods abstractFailMethods would make sense i think... and then we can define it as an interface from the Typescript point of view and be somewhat closer to threeten i think.... 

https://github.com/ThreeTen/threetenbp/blob/master/src/main/java/org/threeten/bp/temporal/Temporal.java

phueper:
Thinking about this a little more i think it makes sense to it like this if it helps typescript support...

@ianp would you mind adding the `minus` functions as well? 

i think then this PR would be complete, and  we can merge it.

phueper:
closing this for now, feel free to reopen if the requested changes have been done

PR #208, opened 2017-12-19
js-joda.d.ts: Add TextStyle enum fields

delego-wlritchi:
TextStyle.FULL and friends currently lead to the following error when using TypeScript:

> error TS2339: Property 'FULL' does not exist on type 'typeof TextStyle'.

This PR adds type definitions for TextStyle's enum members.


phueper:
Sorry we cant accept the PR without tests. Please feel free to reopen again if some test are added.

PR #211, opened 2017-12-27
Introduce static class properties

pithu:
Since static class properties are tc39 stage 3 now https://github.com/tc39/proposal-class-fields and there are esdoc
and babel plugins for it, we could use it to improve the documentation for
static class properties and might think about declaring private class properties with
the hash tag. Both are things that are widely used in threeten.

Unfortunatly the most IDE's do not support the feature yet.

And its not possible, in the most cases, to initialize the static properties right away. The problem with the dependency cycles keeps the same. So we have to stick to the _init functions.


pithu:
Ups, looks like i oversee eslint ... so its still wip

src/LocalDate.js
```diff
@@ -82,6 +82,22 @@ const  DAYS_0000_TO_1970 = (DAYS_PER_CYCLE * 5) - (30 * 365 + 7);
 
 export class LocalDate extends ChronoLocalDate{
 
+    /**
+     * The minimum supported {@link LocalDate}
+     * This could be used by an application as a "far past" date.
+     *
+     * @type {LocalDate}
+     */
+    static MIN;
```
phueper: Hm... where is the definition of `MIN` ... do we still need the `_init` trick? or can we define the statics here ?

pithu: Yes, unfortunatly we still need the `_init` trick. 

The benefit is that we have nicer esdoc with an extra `Static Member Summary` section, with the downside that we have that code without a definition. 

Well, i am not sure, if we should introduce static members unless we solve the dependency cycle thingy...

pithu:
Closed until we find a solution for the resolve issues.

Issue #214, opened 2018-02-15
🔥 Provide esm version


apertureless:
# 🙏 Feature Request

Hey, thanks a lot for this lib.

However I think it would be really awesome, if you could provide an `esm` build, too.
The reason is quite simple. Your source looks very clean and modern. And is using `import` and `export`.  So **imo** this makes it perfect for tree shaking.

## 🕔 Current State

Right now if I am importing 

```js
import { LocalTime } from 'js-joda'
```

the whole `js-joda` lib will end up in my bundle. Because tree shaking is not working.
This is because you're providing only a `umd` build. Which is awesome if you're using it in the browser. However tree shaking is not working in this build.

## 💉 Fix 

Should be quite simple to add a [rollup](https://github.com/rollup/rollup) config to output an ES module and in package.json just add the target in the `module` and `jsnext:main` field

```json
 
"module": "dist/js-joda.esm.js",
 "jsnext:main": "dist/js-joda.esm.js",
```

This way modern bundlers like rollup and webpack 2+ will pickup the esm and be able to tree shake and older bundlers webpack 1, gulp or browser will pickup the `umd` version.

pithu:
Hi @apertureless thanks for reporting and for using js-joda.

I absolutely agree, it would be great to have es module section. I pushed a first draft on a branch `js-joda/js-joda/esm-experimental`

Unfortunately the code in that branch with the es module breaks the webpack 3 build of a project depending on js-joda with a strange error message.

So we still have to investigate in this issue, if you find some obvious problem in the branch please let us know.

pithu:
Short update:
I released version js-joda@v1.7.0 with an esm module build with rollup, but oversee that we still need to transpile some fancy things, like arrow functions, classes, etc. in the esm build.
So there is still some work todo, help is always welcome.
Until than i simply removed the pkg.module prop again (Release @v1.7.1).

apertureless:
Cool. Sorry for the late reply. Somehow did not get a notification.
I will take a look at the rollup file. I think babel is missing there. 

I will provide a PR later. :hand:

pithu:
Finally its there. js-joda has now a pkg.module section with an esm build.

Beside your PR there where some more things to fix: like a node style module.exports in the code and a special babel config for the node test runner. 

I took the opportunity and switched also all other builds to rollup. Just the karma tests are still running with webpack and waiting for migration.

Thx a lot for your hints again. I will close the issue. Please feel free to reopen it again, if there are some issues left.

PR #220, opened 2018-03-02
⚠️ WIP: Rollup config

apertureless:
This is a WIP PR for the rollup / esm build

---

1. Replaced `babel-preset-2015` with `babel-preset-env`
2. Add `rollup-plugin-babel` to rollup config to transpile the source first.

---

## Reasoning

Generally `babel-preset-env` is the way to go, as it respects current browser functions & API and you have better control over the browser support. I added right now `last 2 version` however this can be set more strictly, if you want to support older browsers.

`rollup-plugin-babel` [1](https://github.com/rollup/rollup-plugin-babel#modules) requires that babel does not transform the sources as a module.  Thats why `modules: false` is set. As it defaults to `commonjs` (for node modules) 

I kept the `loose: true` option as you had it in before.

However this still needs testing. And I am not sure, if treeshaking will work, as I am getting a lot of `Circular dependency` warnings.

---

[1] - https://github.com/rollup/rollup-plugin-babel#modules

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,11 +1,14 @@
 {
   "presets": [
-    [
-      "es2015",
-      {
-        "loose": true
-      }
-    ]
+      ["env", {
+          "modules": false,
+          "loose": true,
+          "targets": {
+              "browsers": [
+                  "last 2 versions"
+              ]
+          }
+      }]
   ],
   "comments": false,
   "env": {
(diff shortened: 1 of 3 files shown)


pithu:
Looks like we need two different babel presets, one for the umd build and the other for the ems build.

phueper:
Hi thanks for the PR @apertureless 

I think `babel-preset-env` is great :+1:  ... from what i see at first glance the only difference breaking the build is `modules: false` ... couldn't we move that option to the `rollup.config.js` file ? then we can share the `.babelrc` between `umd` and `ems` build? 
I can't try currently, but i think from the docs that could work?

Cheers, Pattrick

pithu:
see #214

PR #246, opened 2018-07-12
add OffsetClock

henryw374:


henryw374:
got a lint error. now realised there's equals and withZone not implemented on the other clock impls, so I'm doing those

PR #258, opened 2018-08-25
Add DateTimeParseException to typings file

### LINKED ISSUE #257: No TypeScript definition for DateTimeParseException
When trying to do `import { DateTimeParseException } from "js-joda"` in a TypeScript file, I get the following error:

```
[ts] Module '"node_modules/js-joda/dist/js-joda"' has no exported member 'DateTimeParseException'.
```

Looks like the type definition is missing for this exception. Could they be added? Unfortunately, I'm new to TypeScript, and I don't know how to do this myself.

singingwolfboy:
This appears to fix #257. After playing around locally, this change seemed to resolve my problem. No idea if I'm doing it correctly beyond that, though!


singingwolfboy:
@pithu Thanks for the tips! I made the changes you requested, and re-ran the tests -- everything looks good. When I re-ran the tests, it also rebuilt `js-joda.min.js` -- I committed that change too, but in a separate commit so that it's easy to remove if you decide that it doesn't belong in this pull request.

pithu:
Cool thx a lot for contributing. `js-joda.min.js` only needs to be updated before a release, so i just cherry picked the first two commits.

Issue #264, opened 2018-09-22
TypeScript definitions for ZoneId missing `id()`


kenzierocks:
It looks like the `.d.ts` is missing `ZoneId#id()`, defined [here](https://github.com/js-joda/js-joda/blob/master/src/ZoneId.js#L135).

PR #265, opened 2018-10-08
Add `ZoneId#id()` to type definitions

firoxer:
Hi, I noticed #264 and thought it would be a small change to fix it. I corrected both the TypeScript and Flow definitions assuming that the docs about the return type at https://github.com/js-joda/js-joda/blob/2ea179c3588f848b16254db41c19849cc8992e94/src/ZoneId.js#L133 are correct. Hope this is alright.

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -1234,6 +1234,8 @@ declare namespace JSJoda {
 
         hashCode(): number
 
+        id(): string
+
         normalized(): ZoneId
 
         rules(): ZoneRules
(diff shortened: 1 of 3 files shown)


pithu:
Hi @firoxer thanks for using js-joda and for contributing.

Could you add a section to the typescript test file as well `test/typescript_defintions/js-joda-tests.ts` ?

That would be very helpful

firoxer:
Would something like this suffice, appended to the end of the file you mentioned?

```
function test_ZoneId() {
    var id = ZoneId.SYSTEM;

    id.id();
}
```

It's really simple but it seems to be in line with the other examples.

pithu:
Looks good, thats it. What's about renaming var `id` to `zoneId`

firoxer:
Yeah, `zoneId` looks better. I pushed the test to my fork now.

pithu:
Thx a lot!

PR #269, opened 2018-10-21
Edits to public documentation

HerbCaudill:
This PR proposes extensive edits for grammar and clarity to the public site in gh-pages: Mostly to the [cheat sheet](https://js-joda.github.io/js-joda/cheat-sheet.html) and also to the [front page](https://js-joda.github.io/js-joda/). 

The section on time zones has been expanded a bit to clarify the distinction between time zones and offsets, which was confusing to me at first. 

Other changes include:
- Rewriting sentences for grammar and/or clarity
- Consistent use of `code` ticks for inline references to method names etc.
- Consistent capitalization (e.g. JavaScript)
- Consistent spacing (e.g. time zone, not timezone)

I hope this isn't too much at once. Some automated formatting by Prettier makes it look like more than it is.

PR #271, opened 2018-10-24
Edits to documentation for grammar & clarity; consolidation with esdoc

HerbCaudill:
This replaces #269. 

I've taken this a little further and consolidated everything into esdocs. This generates a single public site like this:

image

You can see a live version on my fork: https://herbcaudill.github.io/js-joda/

Besides presenting a single public site that brings together the hand-written manual with the auto-generated reference docs, this should make the workflow around the public site a lot simpler: Just run `npm build-esdocs` before deploying; the site is generated in `/docs`. You can set GitHub pages to serve directly from this directory in the `master` branch, so there's nothing further to do.


phueper:
Hi @HerbCaudill 

wow! ... this looks really great .. .thanks a lot .. i will try to review it in the next few days and talk to @pithu ...

Thanks!

Pattrick

PR #274, opened 2018-10-29
update readme, mark Milestone 3 reached :) 

phueper:
point to the js-joda-locale plugin


phueper:
added/fixed the example in js-joda-locale 2.0.1

PR #276, opened 2018-10-30
Add page on date/time formatting to docs

HerbCaudill:
New page consists of:

- Simple example of using `DateTimeFormatter`
- Explanation of locales & example using prebuilt package
- Example of using built-in `Intl.DateTimeFormat` API to format
- Reference table for Java SimpleDateFormat

Live preview: https://herbcaudill.github.io/js-joda/manual/formatting.html


HerbCaudill:
BTW I ran `build-esdoc` locally and included the output - let me know if it would be better to leave that to your build process.

phueper:
> BTW I ran `build-esdoc` locally and included the output - let me know if it would be better to leave that to your build process.

that is ok, since we want the output in the repo for github pages to pick them up, i think it makes sense like this... 
we never really had a build process for this, it's basically manuall running `build-xxx` and then committing the result :)

HerbCaudill:
yeah, I was wondering about that. Let me see what I can figure out .

docs/class/src/Duration.js~Duration.html
```diff
@@ -2653,7 +2653,7 @@ <h3 data-ice="anchor" id="static-method-parse">
   
   <div data-ice="description"><p>Obtains a <span><a href="class/src/Duration.js~Duration.html">Duration</a></span> from a text string such as <span>PnDTnHnMn.nS</span>.</p>
 <p>This will parse a textual representation of a duration, including the
-string produced by <span><a href="class/src/YearMonth.js~YearMonth.html#instance-method-toString">toString</a></span>. The formats accepted are based
+string produced by <span><a href="class/src/Period.js~Period.html#instance-method-toString">toString</a></span>. The formats accepted are based
```
HerbCaudill: This looks to be an ESDoc bug. Both versions are wrong. The `{@link}` directive in the comments should resolve to `class/src/Duration.js~Duration.html#instance-method-toString`. Instead ESDoc is resolving to the `toString` member of an arbitrary other class.

All the other changes are of this nature.

phueper: Hi, do you have any further info on this? A bug link? Is it known in esdoc? I'd really like to get this pr merged... But I'm unsure wether we should wait for this to be fixed? 🤔

HerbCaudill: It turns out my esdoc build was turning out differently because I was running it from powershell - running it from bash left all those files untouched. 

Note that the pre-existing problem with some `{@link}` directives resolving incorrectly is still there - I'll open a separate issue for that.

HerbCaudill:
@phueper this should be good to go

phueper:
thanks again!

PR #277, opened 2018-10-30
Fix DateTimeFormatter.withLocale() definition (typescript/flow)

HerbCaudill:
`locale` parameter was missing from `DateTimeFormatter.withLocale()` method. 

image


NickDarvey:
Is there any chance of getting this fix released with a new package version? (It looks like [the latest version (1.9.2)](https://www.npmjs.com/package/js-joda/v/1.9.2), was released before this fix was merged.)

PR #291, opened 2019-02-19
Update TypeScript definitions

liamjones:
ZonedDateTime.ofLocal() should allow null as well as ZoneOffset for preferredOffset. This matches the documentation and the behaviour in the function itself


pithu:
Hello @liamjones 

thanks a lot for using js-joda and contributing. There is just a test missing, could you please add a test case for you changes here [test/typescript_defintions/js-joda-tests.ts](https://github.com/js-joda/js-joda/blob/master/test/typescript_defintions/js-joda-tests.ts)

Thanks Philipp

liamjones:
Test added. When doing so I realised that the previous definitions _did_ work, as long as you weren't running tsc with strictNullChecks. I've enabled strict mode completely in test-ts-definitions as the current definitions (with my updates) still pass with all the strict mode options enabled.

I also noticed the test directory was mis-spelt so I corrected that and then had a quick run through for other spelling corrections too.

Issue #254, opened 2018-08-16
Support for no colon in UTC offset


tapatio:
The ISO 8601 standard allows for the UTC offset to be specified without a colon:

`2018-06-12T19:23:26.880+0000`

Ref: [https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC](url)

However, js-joda only supports UTC offsets with a colon:

`2018-06-12T19:23:26.880+00:00`

Can you please add support for UTC offsets without the colon?

Thanks.

pithu:
Hello @tapatio 

thanks for reporting and for using js-joda.

Actually js-joda default formatter is not parsing full featured ISO but more or less [RFC 3339, section 5.6](https://www.ietf.org/rfc/rfc3339.txt).

Nevertheless you can define your own specific parser pattern with js-joda. Here an example that parses an offset without a colon.    

```
const f = DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss.SSSX")
ZonedDateTime.parse("2018-06-12T19:23:26.880+0000", f)
````

You can check the doc for the patterns https://js-joda.github.io/js-joda/esdoc/class/src/format/DateTimeFormatter.js~DateTimeFormatter.html and there are a lot of issues about that on stack overflow.

I hope that helps

tapatio:
Thanks for the prompt reply!

I tried the formatter pattern above that uses "X" as defined here:

X      |     zone-offset 'Z' for zero   |    offset-X    | Z; -08; -0830; -08:30; -083015; -08:30:15;

That works and I am able to parse "2018-06-12T19:23:26.880+0000". However, I am then unable to parse "2018-06-12T19:23:26.880+00:00". I tried "x", and "Z" options and they did not work either. 

Is there a way to configure a formatter to parse both "0000" and "00:00"? 

Thanks.

Issue #249, opened 2018-07-31
Parsing ISO8601 DateTime with Time 24:00:00 throws DateTimeParseException


polmabri:
According to ISO8601 24:00:00 is valid and denotes midnight at the end of a calendar day:

https://en.wikipedia.org/wiki/ISO_8601#Times

```javascript
LocalDateTime.parse("2018-07-31T24:00:00")
```
throws:
```
DateTimeParseException: Text '2018-07-31T24:00:00' could not be parsed: Invalid int value for HourOfDay: 24: 2018-07-24T24:00:00, at index: 0
```

pithu:
Hi @polmabri, 

thanks for using js-joda and your question. Indeed 24:00:00 is part of the ISO spec and can not be parsed neither by js-joda nor by threeten. 

As far as i know, the reason is because its to hard to implement. See this post for further explanation https://stackoverflow.com/a/5388148.

I will close this issue for now, please feel free to reopen it, if you have further remarks.

Issue #240, opened 2018-06-26
OffsetTime missing


henryw374:
Hi, 

Firstly thanks for js-joda! I would like to know if you have plans to implement OffsetTime? 

I found another ticket discussing OffsetDateTime https://github.com/js-joda/js-joda/issues/165 and with that one it seems to make sense that the functionality already exists. 

Thanks,
Henry

pithu:
Hello @henryw374 

Unfortunately the situation did not change since #165, we have no plans to implement OffsetTime or OffsetDateTime, but we would accept a PR.

henryw374:
Ok thanks... Might try that if I have time

henryw374:
work started in this fork https://github.com/henryw374/js-joda

Issue #181, opened 2017-08-22
Parsing basic (w/o delimiters) ISO:8601?


slavafomin:
Hello!

Thank you for this great library!

However, how do I parse basic (w/o delimiters) ISO:8601 string using it? I have a time string, which looks like this: `20170822T194135+00`. But js-joda is failing to parse it through `ZonedDateTime.parse()` for some reason.

What would be a correct way to parse it then?

As far as this is a correct ISO:8601 string, I was expecting js-joda to parse it normally.

Thank you!

slavafomin:
Also, when I try to parse it using custom format:

`const time = ZonedDateTime.parse(timeString, DateTimeFormatter.ofPattern(`yyyyMMdd'T'HHmmssZ`));`

I'm getting error on column `15` (that's where the timezone is encountered). Why is that?

pithu:
Hello @slavafomin 

js-joda uses predefined formater like specified here https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

To parse the format in your example you have to define a special formatter pattern: 
```
const f = DateTimeFormatter.ofPattern("uuuuMMdd'T'HHmmssX")
ZonedDateTime.parse('20170430T194030+00', f).toString()
```
Hope that helps.

slavafomin:
Thank you @pithu for the example, I'm using a quite similar approach right now. However, shouldn't the library be able to parse any valid ISO:8601 string out of the box without defining any "custom" formats? I think this is the purpose of standards in the first place to make data interchangeable and easy to parse.

I've checked that `moment.js` understands this format without any issues out of the box.

pithu:
Well, we are quit strict in following the threeten implementation. And the threeten philosophy is to be strict and explicit.

jsumners:
I have to agree with @slavafomin, being able to natively parse any ISO 8601 compliant string should be an expected feature in any time/date library. I came to the issues specifically to see if this had already been raised.

pithu:
Well, most of the moment code is for parsing and formatting time. The ISO 8601 has quite a lot of options to format time data. Parsing all possible options would make parsing slower and lead to more code, just to handle some rare use case. That's one of the reasons why we have no plans to change it.

[...]

PR #295, opened 2019-03-05
Explicit callout of not supporting all of ISO8601

gjbianco:
While I understand why (and even agree with) js-joda doesn't fully support ISO8601, the README mentions ISO8601 11 times and never explicitly mentions that the full standard is not supported.

It also looks like I'm not the only one that's been hit by this either:

#249 , #254 , #240 , #181 

Eventually, it would be really beneficial to enumerate which parts of ISO8601 are or are not supported.


pithu:
@phueper do you have an opinion ?

gjbianco:
I think that formulation is still vague about what is _not_ supported. Personally, I think the most clear wording explicitly describes that not _all_ of ISO8601 is supported.

phueper:
hmmm... i think maybe it's better to describe what we support instead of stating what we do not support? would that be possible? e. g. describing the formats that are supported out of the box...

gjbianco:
@phueper Definitely, but I don't think there is an explicit enough term for the exact parts of ISO 8601 that js-joda supports.

pithu:
Well as i mentioned above, the build in format/ parse functions do support `RFC 3339, section 5.6`, isn't that very explicit ?

gjbianco:
"Note well". I can change it to just "note: ..." if you prefer.

https://en.wikipedia.org/wiki/Nota_bene

pithu:
In my opinion, thats quite to much detail information at this prominent place, what about 
`based on the ISO calendar system, which is the de facto world calendar following the proleptic Gregorian rules.` instead of `based on the ISO8601 calendar.`. Might it make more clear

And then add the reference to `RFC 3339, section 5.6` to the according `format` and `parse` methods of `LocalDate`, etc.

gjbianco:
The whole _point_ of this PR is to make it more clear that the library (intentionally) does not support all of ISO8601. It is apparently not clear _enough_ since there are at least 4 documented cases of developers running into this issue.

I do not want to bury this in the documentation. If somebody is looking for a library to support ISO8601, it should be prominent that this library only supports a subset. Even if that's likely all they need, it should be as clear as possible _before_ they start digging into particular function docs.

Also, I didn't even add "based on the ISO8601 calendar" -- it was part of the original README.

pithu:
Find my changes in the fix, fill free to open the issue again if i missed something

PR #296, opened 2019-03-14
add type definition for ChronoUnit.NANOS

yobryon:
add missing definition to ChronoUnit class for the NANOS value that is in the underlying lib.


pithu:
Hello @yobryon 

thanks a lot for using js-joda and contributing. There is only a test missing, could you please add a test case for you changes here test/typescript_defintions/js-joda-tests.ts

Thanks Philipp

yobryon:
Added a rudimentary test in the style of the ChronoUnit.MILLIS tests.  If you'd like a more complete functional test, I can work on that!

PR #298, opened 2019-03-25
Adding private constructors and abstract modifiers to TypeScript definitions

### LINKED ISSUE #279: Reconsider add again private constructors to typescript type definitions
Many constructor in this projects are private. Since they are no intended to be used from clients of the library they should be marked as such in .d.ts files. I think this used to be the case (see ecba785a495a3beec660a4c3208561a5b5d6bb1a and [#134-comment](https://github.com/js-joda/js-joda/issues/134#issuecomment-280951131)) but then they were removed.

If a class lacks a `constructor` declaration the class is still constructible (with a default, zero-argument constructor). So, the next snippet will fail at runtime but not at type-checking

```ts
const x = new Instant(); // Compiler is happy
```

This is wrong since, a) the constructor (right now) expects two arguments, and b) it is not expected to instantiate the class like that. When I'm using TypeScript I expect to be notified at type-checking of this misuses, right now and while I'm learning the API this is confusing.

Adding a `private constructor` declaration will prevent this, even if the constructor has no parameters:

```ts
// Somewhere inside the type definition
class Instant extends Temporal {
    // ...
    private constructor();
    // ...
}

const x = new Instant(); // ERROR: Constructor of class 'Instant' is private and only accessible within the class declaration
```

So, please reconsider to add those declarations again.

InExtremaRes:
This PR is an attempt to improve the TypeScript types definitions an bring them closer to implementation. Closes #279.

As discussed in #279, the constructor of many classes are implementation details, but since constructors were previously removed from TS defs for classes they were all _newable_ with a zero-parameter constructor. Following the JSDoc annotations in source code and the docs of the [ThreeTen backport](https://www.threeten.org/threetenbp/apidocs/index.html) I made the following changes:

- All enums have got now a private constructor so they are not instantiable nor extensible.
- Most of the classes marked as `final` in ThreeTenBP (for instance `Instant` or `ValueRange`) now have got a private constructor. And exception was made for `DateTimeFormatterBuilder` since the that constructor seems to be public.
- Some classes are now marked as `abstract`.

**Other changes I made:**

- Enums values added to `TextStyle`.
- Class `OffsetDateTime` and method `ZonedDateTime.toOffsetDateTime()` were both removed since are not implemented in JS.
- Since `Chronology` doesn't exist as a class in JS it is now an alias type for `IsoChronology`.
- `TemporalField` was previously declared as an `interface` but is a `class` in JS, so it is now declared as an abstract class.
- `ZoneRules` were missing a lot of abstract methods that are now declared.
- Added the definition type for `ZoneOffsetTransition`.

**Some questions:**

- Is `MathUtil` meant to be used by users of the library? It has type definitions but it looks as an internal utility class to me.
- `TemporalAccesor`, `ChronoLocalDate` and maybe other classes are interfaces or abstract classes in ThreeTenBP, but here are implemented as full classes with concrete methods. Why? I didn't type those as `abstract` for that reason. 

**A little problem:**

Js-Joda simulates interfaces as classes where every method has an `abstractMethodFail` assertion. It has a drawback: a class "implementing" that interface must use `extends`, but JS has single inheritance so a class cannot really "implement" more than one.

In TypeScript there is an `interface` keyword that creates a new type that can be implemented by classes and a class can implements many of them. However we cannot use `interfaces` in type definitions since they are elided when transpiling to JS (are just compile time artifacts) and classes who implemented those doesn't have an `extend` clause (except that the class extends some base class in addition to implement some interfaces). Then any method that expect and object to be an instance of any of those "interfaces" (at runtime) will break.

I suppose that is not a problem for most users of the library since very few, I guess, are implementing those interfaces.

### DIFF
--- a/dist/js-joda.d.ts
+++ b/dist/js-joda.d.ts
@@ -1364,20 +1364,84 @@ declare namespace JSJoda {
         rules(): ZoneRules
     }
 
+    class ZoneOffsetTransition {
+        static of(transition: LocalDateTime, offsetBefore: ZoneOffset, offsetAfter: ZoneOffset): ZoneOffsetTransition
+
+        private constructor()
+
+        instant(): Instant
+
+        toEpochSecond(): number
+
+        dateTimeBefore(): LocalDateTime
+
+        dateTimeAfter(): LocalDateTime
+
+        offsetBefore(): ZoneOffset
+
+        offsetAfter(): ZoneOffset
+
+        duration(): Duration
+
+        durationSeconds(): number
+
+        isGap(): boolean
+
+        isOverlap(): boolean
+
+        isValidOffset(offset: ZoneOffset): boolean
+
+        validOffsets(): ZoneOffset[]
+
+        compareTo(transition: ZoneOffsetTransition): number
+
+        equals(other: any): boolean
+
+        hashCode(): number
+
+        toString(): string
+    }
+
+    interface ZoneOffsetTransitionRule {
+        // TODO: Not implemented yet
+    }
+
     abstract class ZoneRules {
         static of(offest: ZoneOffset): ZoneRules
 
-        isFixedOffset(): boolean
-
-        isValidOffset(localDateTime: LocalDateTime, offset: ZoneOffset): boolean
+        abstract isFixedOffset(): boolean
 
         offset(instantOrLocalDateTime: Instant|LocalDateTime): ZoneOffset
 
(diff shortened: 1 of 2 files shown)


pithu:
Thx a lot, looks good. 

Yes, MathUtil meant to be for internal use only. We just need it in some other js-joda libs, like js-joda-locale. 

As far as i remember there are some abstract classes around TemporalAccesor and ChronoLocalDate, thats needs some stuff to be implemented. They should probably be declared as abstract.

PR #313, opened 2019-05-17
TypeScript: Add type definition for Instant.atZone

teozkr:
This was already [documented](https://js-joda.github.io/js-joda/class/src/Instant.js~Instant.html#instance-method-atZone), but not part of the type definitions.


pithu:
Hello @teozkr 

thanks for using js-joda and contributing. 

There is a test missing, can you please add a test case for you changes here test/typescript_defintions/js-joda-tests.ts

Thanks

PR #339, opened 2019-07-31
Add TypeScript declarations for extras package

### LINKED ISSUE #336: No types available for extras
I'm going to produce my own types and would consider contributing to DefinitelyTyped, but seems like there are .d.ts files elsewhere in this repository, hence this issue report. Are these existing types generated automatically in some way?

brandonbloom:
Fixes #336

### DIFF
--- a/packages/extra/dist/js-joda-extra.d.ts
+++ b/packages/extra/dist/js-joda-extra.d.ts
@@ -1,0 +1,51 @@
+declare module '@js-joda/extra' {
+  import { Duration, Instant } from '@js-joda/core';
+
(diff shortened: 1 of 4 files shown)


pithu:
Hi @brandonbloom,

great thanks a lot, could you also add some tests, that ensure that the definitions are correct, similar to the core package ? see
https://github.com/js-joda/js-joda/blob/master/packages/core/package.json#L27

brandonbloom:
Right on. Added!

pithu:
thx a lot!

PR #204, opened 2017-12-11
Add some missing methods to "Year"

benny-medflyt:

PR #353, opened 2019-10-14
Improve of the TS declarations

### LINKED ISSUE #347: `JSJoda.Instant is not assignable to parameter of type 'import (.../core/dist/js-joda).Instant`
the new published extra package seems to be missing Instant definitions from core, so its unusable.

### LINKED ISSUE #349: [core] Missing static methods in TypeScript definitions
Some classes has missing or wrong static methods definitions. For instance, [definitions for LocalDateTime](https://github.com/js-joda/js-joda/blob/master/packages/core/dist/js-joda.d.ts#L923) says that all parameters of `of()` are optional but that isn't true. Also there is no definition for `ofNumbers`, `ofDateAndTime`, etc. Same happens in other classes.

I could give that a try and send a PR if none else can. All those static methods are public (`ofNumbers` and so on)? Are public all of those `of2`, `of8`, etc?

InExtremaRes:
This PR fix the TypeScript declarations, improving a little the type safety and hiding some non-public functions.

- Some parameters and return values previously typed as `any` now get a more specific type. There is more work to do here, but this change, even if small, is an improvement.
- ~The previous point is more noticeable in the `equals` methods that are now typed to only accept the same instance than the class they are declared. I am aware that `equals` accepts any value and returns `false` when the type doesn't match, but I think most of the time it is a non-intentional developer mistake to compare values of different types; in the rare cases it is the developer intention to do that (for example to pick a value from an array of `Temporal`), the "unsafe" comparison can be expressed explicitly with something like `value.equals(other as any)` or similar.~

    **Edit:** As requested, now every `equal` method has a single parameter of type `any`.

- Some static methods were not intended to be public so they're removed from the declarations. Some others got a parameter name fixed or were marked as optional.
- The declaration file is now explicitly declared as an ES module. This is perhaps the most important change of this PR so I will explain in detail below.
- Declaration file for the _extra_ package is now an ES module also and uses the types of _core_ correctly.
- Some new methods are typed, like `Year`.
- File is also formatted, mostly automatically.

The way the file was written, with a `declare namespace JSJoda` and an `export = JSJoda`, is a common pattern for CommonJS modules. It works but is not very precise since the distributed .js file is build as an ES module and doesn't include a `default export`.  For instance TypeScript will not fail at compile time if we import the library like this (assuming `esModuleInterop: true`):

```ts
import joda from '@js-joda/core';
```

Even more, TS suggests to convert this, which is valid:

```ts
import * as joda from '@js-joda/core';
```

to the former type of import.

The file is now using a more common pattern for modules that only have named exports and the `__esModule: true` property informs TS that this is an ES module so it cannot be converted to a default import even if `esModuleInterop` is set. That property is automatically defined by Babel.

I make every change in a different commit so it would be easy to split this in two or more different PRs if you are happy with only some of the changes.

Closes #349, closed #347


InExtremaRes:
> Hi @InExtremaRes,
> 
> thanks a lot for the PR, looks like great work. I take a first look, looks good, a lot of clean up.
> 
> I am not an expert with typescript, so i have no opinion on the declaration as esmodule. but i would merge it, as long nobody complains.

Thank you!

> But i disagree with the equals change. equals should take anything, that how it's declared in threeten and in java and that's how in common equals works.

I know that it is common in Java, but in TypeScript it is far less common, at least from a type perspective. Just see this simple code:

```ts
let x = 123;
let y = '123';
if (x === y) {} // Error here!
```

That comparison gives this error at compile time:

```plain
This condition will always return 'false' since the types 'number' and 'string' have no overlap. ts(2367)
```
Similar in this case:

```ts
const array = [1, 2, 3];
if (array.includes('a')) {} // Error here!
```
That reports this error:

```plain
Argument of type '"a"' is not assignable to parameter of type 'number'. ts(2345)
```

As you can see, although both expressions are well defined (both just give `false`), TypeScript complains since its philosophy is to catch the majority of mistakes at compile time, even if those are not strictly errors: in the examples (there are plenty of those, these are just a couple) it is very likely that the developer is making a mistake, as picking the wrong argument or property, and that error could be very difficult to find in a more complex code. TypeScript will catch the error immediately and that is what TS user usually expect from a library.

If a have the code:

```ts
function foo(arg1: LocalDate, arg2: LocalDateTime) {
  if (arg1.equals(arg2)) {
    // make something utterly important
  }
}
```

I will expect that to fail at compile time since the comparison obviously have no sense, it will not convert one to the other and instead always returns `false`. It is not that the code _cannot_ work but that it is _unsound_. Why should anyone write a code like that save as an unintended bug?

pithu:
Well, we have been always very strict in following the threeten api as is.

Beside that, i would expect that the typescript api is as close a possible to the javascript api. Therefore i would expect the that following code give the same result in javascript as in typescript.

```
LocalDateTime,now().equals(LocalDate.now()) // expect to be false and not fail on compile time
```

Its true, that use cases for that might be rare, but i could imagine that it could be useful in some cases, like if you have a list of objects of type `Temporal` that you want to compare:

```
for (temporal of listOfTemporals) {
	if(localDateTime.equals(temporal)) {
		// do sommething
	}
}
```

Claudisimo:
`LocalDateTime,now().equals(LocalDate.now())`

Is something that you should expect typescript  to warn you and fail at compilation time, since TS advocates for a stronger use of types. If you do end up in a situation like the example shown above typescript should show the error and you'll had to make the right assertions inside that 'for' loop or allow .equals method to transform the types it can when it makes sense (like LocalDateTime to a LocalDate), however that would not be "equality" in the strict sense.

I understand the idea of triying to follow the threeten api as close as possible, but the idea of typescript is not allow those kind of things to happen, so if this are TS declarations I would think that rule takes precedence. Common equals in typescript just don't work like java ones or others.

Sorry for my horrible english.

InExtremaRes:
I understand your position, just let me argue one last time :)

> Well, we have been always very strict in following the threeten api as is.

I understand that, but it is important to notice that `.equals(Object o)` is very idiomatic in Java since it is inherited from `Object` with that signature. That doesn't happen exactly in that way in JavaScript. Using `equals` is just a convention people have to learn and there is no preconceptions about its semantics as in Java.

> Beside that, i would expect that the typescript api is as close a possible to the javascript api. Therefore i would expect the that following code give the same result in javascript as in typescript.
> 
> ```
> LocalDateTime,now().equals(LocalDate.now()) // expect to be false and not fail on compile time
> ```

Here is where JavaScript and TypeScript start to diverge, slightly. Every TypeScript programmer knows that it is more strict than JS, but that is exactly the point of using it. JavaScript has a lot of oddities and TypeScript is there to say "look, that thing you are trying to do, is very very likely that doesn't do what you are thinking"; that is exactly what I would expect to happen in that last snippet and I bet the same for other programmers using TS. That is because TypeScript is not just type annotations but is very much a programming language in its own right, just very close to JavaScript. It builds the (eco)system around the types so the types can guide your development.

If I see a library written for JavaScript but with TypeScript declarations I would expect its types to follow the TypeScript conventions nonetheless, otherwise I'd be using just JavaScript. It is very common that a library is more permissive in JS but takes advantage of the type system on types declarations.

> Its true, that use cases for that might be rare, but i could imagine that it could be useful in some cases, like if you have a list of objects of type `Temporal` that you want to compare:
> 
> ```
> for (temporal of listOfTemporals) {
> 	if(localDateTime.equals(temporal)) {
> 		// do sommething
> 	}
> }
> ```

In a case like that (also not that usual) it is just a matter of doing:

 ```ts
for (temporal of listOfTemporals) {
  if(localDateTime.equals(temporal as LocalDateTime)) {
    // do sommething
  }
}
 ```

The `as T` is called a _type assertion_ and is not a casting, it incurs no penalty at runtime, and no conversion is performed, it is just the programmer saying "dear TS, I know that this types can be compared in this way", and I think that is a good thing because it makes your intentions explicit. TypeScript programmers knows how to make a type assertion for edge cases like this. (As a side note, this kind of assertion still have some type-safety because TS will check if the types are somehow related, it is possible to say "ey, trust me, I'm a programmer" and make an assertion between types totally unrelated, but it is not the case here).

Finally, as you can see in the diff some of the `.equals` where already typed to accept only its own instance type, I just homogenize the rest that where typed as `any`. If people were using those types for a while now without complains I guess there is no problem in change the other ones.

phueper:
Hi

I'm not much of a TS developer, but I tend to agree with @pithu and would like to keep the equals accepting `any`. But I think they should be consistent in the ts definitions not some accepting any and others not.
In any case, changing this would mean a breaking api change for ts users and require us to do a major version increment? 🤔

InExtremaRes:
Ok then. I will rebase and push again soon.

> In any case, changing this would mean a breaking api change for ts users and require us to do a major version increment?

I am not sure. The change for `equals` accepting `any` everywhere shouldn't be a breaking change since it makes the types more lax. The change from a CommonJS to an ES module _could_ be, but I don't so; however I cannot assure that since, even when we are using these types in our projects for several days now, we are working in Node.JS and not in browser. The tests for TS still pass without changes so I guess is safe.

pithu:
Ok fine, i have really a strong opinion here, the api should be the same across all languages.

I saw that the javascript docu is also wrong for some equals method, instead of using `*` it refers to to specific class, so we should fix that as well.

Its definitely a major release, because @InExtremaRes  removed things like `MonthDay.ofMonthNumber`, etc. 

So it would be great if there are some more things to clean up, we could release it all in one step.

InExtremaRes:
@pithu I have a lot of work right now, but in the next days I'll address the request and will look what else to cleanup/improve. I think `@js-joda/extra` should be changed to an ES module types as well.

Any suggestion about what else should I look?

pithu:
> Any suggestion about what else should I look?

There have been some typescript definitions PR's rejected by us, because of missing tests. I just looked through the PR's and could find only this one #204. That's the only thing that i can remember.

InExtremaRes:
@pithu @phueper The PR is not complete, there is some more cleanup I'd like to do, but I'd like if you can review the changes already made and give me some feedback about the progress.

pithu:
> @pithu @phueper The PR is not complete, there is some more cleanup I'd like to do, but I'd like if you can review the changes already made and give me some feedback about the progress.

Looks good so far, great work !

InExtremaRes:
I rebased onto master so there is no conflicts. It's is now ready for a full review and eventually be merged.

InExtremaRes:
> And / or it would help, if there are some more tests, like for the generic TemporalQuery class i couldn't find any tests.

It happens that I'm not sure how to test it exactly. I mean, a call to `temporal.query(aQuery)` should return the same as `aQuery.queryFrom(temporal)`, which is now enforced by the type parameter `R`

```ts
abstract class TemporalAccesor {
    query<R>(query: TemporalQuery<R>): R;
}
```

So if I do

```ts
const temporal = localDateTime.query(TemporalQueries.localTime());
temporal.dayOfMonth(); // Error: 'dayOfMonth' does not exist on type 'LocalTime'
```
I get the error since TS can infer the type of the resulting `Temporal` from the `TemporalQuery`; before it'd be of type `any` that allows anything. And that is the problem: the test infrastructure for TS declarations does not allow to test an error so I don't know how to test the fact that object is now of the right type rather than `any`. The little utility `expectType<T>()` I made some time ago won't work since `any` is assignable to every type and the test will pass with the code before the change, without the generic. If someone else has an idea to properly test this it would be great!

BTW, I'm noticing that all the types around the `TemporalQuery` thing can be improved a little.

Is every `TemporalQuery` allowed to return `null`? I've noticed that those in `TemporalQueries` returns `null` but `Instant.from` (as an example) throws an exception instead. How could I know which ones return `null` rather than throw? Type those as possibly returning `null` could be very helpful to TS users.

I also notice that some classes have a `FROM` static property I missed before, should I type those as well?

pithu:
I would say, its enough to have at least every method called once in the typescript definition tests, without further functional tests. That ensure, that they don't fail at compile and runtime, that should be enough.

phueper:
I agree with @pithu.. The idea of the TS "tests" was just a file that calls the methods we declared and make sure that tsc does not complain...

InExtremaRes:
@pithu @phueper, done!

I also add types for the `.FROM` static references.

packages/core/test/typescript_definitions/js-joda-tests.ts
```diff
@@ -732,3 +727,40 @@ function test_Temporal() {
...
+    const localDate = temporal3.query(TemporalQueries.localDate());
+    if (localDate) {
+        localDate.atStartOfDay();
+    }
+
+    const zoneId = temporal4.query(TemporalQueries.zoneId());
+    if (zoneId) {
+        temporal3.atZone(zoneId);
+    }
+
```
pithu: That looks strange, looks like there is null returned and the tests in the if statement are not executed.

I played around a bit, whats about something like this ?
```
    const instant: Instant = Instant.now();
    const localDateTime: LocalDateTime = LocalDateTime.now();
    const zonedDateTime: ZonedDateTime = ZonedDateTime.now();


    const unit = instant.query(TemporalQueries.precision());
    expectType<TemporalUnit>(unit);

    const localDate = localDateTime.query(TemporalQueries.localDate());
    expectType<LocalDate>(localDate);

    const zoneId = zonedDateTime.query(TemporalQueries.zoneId());
    expectType<ZoneId>(zoneId);
```

InExtremaRes: Well, as you mentioned before the file is never executed, instead it is just checked by TS so every path is analyzed by the compiler.

Right now you don't have the `strictNullChecks` flag enabled, if it would then TS will notice that the type is possible `null` but it'll narrow the type automatically inside the `if` block at compile time since the control flow analysis "knows" the only way to enter inside that block is when `unit` is not null. I made the test like that since was more natural to me, is more similar to the way I think users will use it and it will not break in case you should decide to enable `strictNullChecks` in the future. The compiler still checks the path inside the `if`, the only difference is that it knows there the type cannot be `null`.

If you still prefer to me to change it I have no problem!

pithu: To be honest, i would prefer to change it. Like you did its quite hard to read in my opinion and it doesn't fit to the patterns we use in the rest of the typscript test.

InExtremaRes: Sure, you're the boss :wink:

InExtremaRes: At the end you do have `strictNullChecks` enabled so I have to assert the exact type outside the `if` block. But change is done.

pithu:
thanks a lot, well done

Issue #355, opened 2019-10-21
Question: When will there be a new major release and what can be changed?


InExtremaRes:
As you know I'm working on improving the TypeScript declarations. While researching in the implementations files and java.time I've found several inconsistencies, not only in types declarations but also between different parts of the codebase and between the library and the ThreeTen API.

Just to name a few:

- `TemporalField` has no abstract methods even when the ThreeTen API has some.
- `IsoFields` has some static instances of class `Field` that extends `TemporalField`, but it has the method `getDisplayName` while `ChronoField` has the method `displayName` instead.
- Every `Temporal` object (I think) should have an `equals` method (like `LocalDate`, `Instant`, etc.) but neither `Temporal` nor `TemporalAccessor` has such abstract method so it is not safe to call `equals` in an abstract `Temporal`.
- This library tries to simulate interfaces with classes, but as I commented in #298 this is not the same. Since JavaScript has only single inheritance you can only extends from one class; the API follows (at some extent) the Interface Segregation Principle and there are cases when there's desirable to "implement" more than one interface, but that doesn't work with classes (at least no without manipulate the prototype chain in some way). The basecode has a lot of `requireInstance` calls for those "abstract classes" what forces users of the library to extends even when there is no behaviour to inherit, preventing them to extend from another, more useful base class.

And there are many more. These issues makes the library difficult to type in TypeScript and a little cumbersome to extends or use in more advanced scenarios.

I'd be very happy to help improving this library, not only with TS types but with patches to the code as well. However I don't understand exactly the vision of the library and the rationale behind some decisions made. I am not always sure if start making some non-trivial changes.

I don't know if there are some other channel to discuss these kinds of topics, but in the meanwhile I'll leave here some questions:

- **When is a major release planned? How much changes can be made?**
- **Is the goal of the library to conform to the ThreeTen API (when it makes sense)?**
- **Can we take away some of the "abstract classes" and simulate interfaces by other means?**
- **Is it OK for me to send a PR making methods names consistent?**

pithu:
> When is a major release planned? How much changes can be made?

No plans, we can release whenever it makes sense or on request. Before a major release its nice to collect some issues.

> Is the goal of the library to conform to the ThreeTen API (when it makes sense)?

Simply yes, in doubt we follow ThreeTen

> Can we take away some of the "abstract classes" and simulate interfaces by other means?

Not sure what you are talking about here, could you give some pseudo code examples ?

[...]

PR #361, opened 2019-10-28
Add missing methods to `Temporal` interface

### LINKED ISSUE #360: `Temporal` interface is lacking methods
The [`Temporal`](https://github.com/js-joda/js-joda/blob/bcc2f3a1cb693b52cb3231787456d4167fbbeca9/packages/core/src/temporal/Temporal.js) interface currently does not define any own methods. In contrast the [ThreeTen/java.time API](https://docs.oracle.com/javase/8/docs/api/java/time/temporal/Temporal.html) specifies methods like `plus`, `minus`, etc.

Without those methods it seems to be impossible to implement for example the `addTo` and `subtractFrom` methods of `TemporalAmount` interface as they need to be able to do a meaningful calculation with the `Temporal` they get.

For example the [implementation of `addTo` in the `Period` class](https://github.com/js-joda/js-joda/blob/bcc2f3a1cb693b52cb3231787456d4167fbbeca9/packages/core/src/Period.js#L792-L807) just uses a `plus` method of the `Temporal` it gets, but according to the defined types this would be a type error.

thrucker:
The `Temporal` interface was missing some methods which are necessary to work with `Temporal` objects. This resolves #360.

This PR adds the abstract methods defined in the [`java.time` API for `Temporal`](https://docs.oracle.com/javase/8/docs/api/java/time/temporal/Temporal.html). Additionally, to be more consistent with the `java.time` API, the superclass of some classes has been adjusted from `Temporal` to `TemporalAccessor`

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -11,9 +11,17 @@ declare namespace JSJoda {
     abstract class Temporal extends TemporalAccessor {
       isSupported(unit: TemporalUnit): boolean
 
+      minus(amountToSubtract: number, unit: TemporalUnit): Temporal
+
+      minus(amount: TemporalAmount): Temporal
+
       plus(amountToAdd: number, unit: TemporalUnit): Temporal
 
-      until(endExclusive: Temporal, unit: TemporalUnit): number
+      plus(amount: TemporalAmount): Temporal
+
+      until(endTemporal: Temporal, unit: TemporalUnit): number
+
+      with(adjuster: TemporalAdjuster): Temporal
 
       with(field: TemporalField, newValue: number): Temporal
     }
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -9,6 +9,13 @@ declare namespace JSJoda {
     }
 
     abstract class Temporal extends TemporalAccessor {
+      isSupported(unit: TemporalUnit): boolean
+
+      plus(amountToAdd: number, unit: TemporalUnit): Temporal
+
+      until(endExclusive: Temporal, unit: TemporalUnit): number
+
+      with(field: TemporalField, newValue: number): Temporal
     }
 
     abstract class Clock {
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -38,7 +38,7 @@ declare namespace JSJoda {
         abstract equals(other: any): boolean        
     }
 
-    class DayOfWeek extends Temporal {
+    class DayOfWeek extends TemporalAccessor {
         static MONDAY: DayOfWeek
         static TUESDAY: DayOfWeek
         static WEDNESDAY: DayOfWeek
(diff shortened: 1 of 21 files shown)


phueper:
@thrucker

Regarding the java.time api... I know that it is probably the most widely used, but due to Java using a GPL license, we decided to follow the threetenbp api (we even think about renaming to threetenjs) ... I suppose they are the same in this case, but could you double check, please?

thrucker:
I checked against the threetenbp API. The changed superclasses for `DayOfWeek`, `Month` and `MonthDay` are also correct for that API.

But I realized that some overloaded methods are still missing in the `Temporal` interface. I'll add those as well.

pithu:
Hi @thrucker,

Thanks a lot for contributing! 👍 I just had a short look at it yet:
- Would be great if you can add some tests for the typescript definitions and for the added functionality
- Is it somehow possible to implement more then one Interface with typescript ? Like DayOfWeek implements TemporalAccessor, TemporalAdjuster. Currently just one Interface is extended, what if we need to add the other one later on ?
- Did you had a look at #353. Are there overlaps or conflicts with this PR?

I will do a full code review as soon as possible.

InExtremaRes:
Hi @pithu.

This changes will conflict with #353 syntactically since I change the whole file. But if this changes are merged I'll be happy to solve #353 (which is conflicting now with the last PR merged, I'll fix that pretty soon).

TypeScript allows to implement more than one interface, but `Temporal` or `TemporalAdjuster` are not interfaces but classes; TS being a superset of JS doesn't allow a class to extends more than one. It is possible to change the types so they are interface in TS but the library assert for inheritance many times.

Hi @phueper.

> [...] we decided to follow the threetenbp api (we even think about renaming to threetenjs) [...]

Could you please explain this a little deeper?

thrucker:
@InExtremaRes, @pithu  I'll try to keep the changes in the typescript definition at a minimum. Maybe I should only add those new methods to the `Temporal` interface and nothing else. Regarding multiple interfaces: As @InExtremaRes says, it's possible but requires proper interfaces instead of classes in the TS definitions. I suggest to do a general overhaul of the TS definitions either in #353 or a separate issue.

InExtremaRes:
@thrucker:
> @InExtremaRes, @pithu I'll try to keep the changes in the typescript definition at a minimum. Maybe I should only add those new methods to the `Temporal` interface and nothing else.

I'm ok with changes you made and is no trouble to me fix #353 onto this afterwards. My only note is about the overloads definition for `plus` and `with`. Did you left `minus` out on purpose?

> Regarding multiple interfaces: As @InExtremaRes says, it's possible but requires proper interfaces instead of classes in the TS definitions. I suggest to do a general overhaul of the TS definitions either in #353 or a separate issue.

I opened an issue recently and in [this comment](https://github.com/js-joda/js-joda/issues/355#issuecomment-546444794) I made a proposal; I'd like to hear your thoughts!

phueper:
> 
> > [...] we decided to follow the threetenbp api (we even think about renaming to threetenjs) [...]
> 
> Could you please explain this a little deeper?

I'm unsure how to explain this deeper... I think it also mentioned in the Readme... The Java api is licensed under a GPL license, we do not want to publish a js license under the GPL because we think it is not very nicely applicable to a js library, thus we decided that our base api is the threetenbp api which is very similar but not always the same as java.time... Qe thought about renaming js-joda to threetenjs for this reason to make it clearer what the base is (since the joda api itself is a little different as well) , but haven't decided if we want to do so. 

Not sure what else I could explain? 

Regards, Pattrick

InExtremaRes:
@phueper Thank you! It is very clear now.

PR #363, opened 2019-11-09
Remove `withTemporalAdjuster` and others non-public methods from TS declaration

InExtremaRes:
This PR remove some methods like `withTemporalAdjuster`, `withFieldAndValue` and `atStartOfDayWithZone` that shouldn't be exposed in the public API.


pithu:
What do think, time to make a release or will more cleanup follow?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

InExtremaRes:
@pithu: I don't find more to clean up right now so I think is good time for a release :)

PR #373, opened 2020-01-06
Add Typescript for IsoChronology.INSTANCE

cranberyxl:
Allow this to pass Typescript type checking:

```ts
const timestamptzFormatter = new DateTimeFormatterBuilder()
  .parseCaseInsensitive()
  .append(timestampFormatter)
  .appendOffset('+HH:mm', 'Z')
  .toFormatter(ResolverStyle.STRICT)
  .withChronology(IsoChronology.INSTANCE);
```

While this is created during the `_init` it is not exposed via typescript.

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -1091,6 +1091,8 @@ export interface Locale {
 }
 
 export abstract class IsoChronology {
+    static INSTANCE: IsoChronology;
+
     static isLeapYear(prolepticYear: number): boolean;
 
     private constructor();


pithu:
Hello @cranberyxl, 

that feels wrong somehow, i would guess, we don't have to expose the `INSTANCE` property. There are no chronologies beside ISO implemented yet. 

`DateTimeFormatter` defaults to `IsoChronology.INSTANCE`, so i would say you can omit `.withChronology(IsoChronology.INSTANCE)` and get the same result.

cranberyxl:
Should `INSTANCE` be private or protected then? It is there and the constructor is private.

pithu:
Internally it must be public, but it should not be exported at all.

I close it for now, feel free to reopen it if you have further questions. 

That should do the trick

```
const timestamptzFormatter = new DateTimeFormatterBuilder()
  .parseCaseInsensitive()
  .append(timestampFormatter)
  .appendOffset('+HH:mm', 'Z')
  .toFormatter(ResolverStyle.STRICT);
```

PR #379, opened 2020-02-05
changed bitwise or to logical or

bowersj:
To allow for safer parsing of integers, in JavaScript, the bitwise or (`|`) operator was switched to the logical or (`||`) operator.

Although this is different from the implementation in the threetenbp -https://github.com/ThreeTen/threetenbp/blob/master/src/main/java/org/threeten/bp/Period.java#L348 repository this change makes more sense in JavaScript since there is only the concept of a double.

### DIFF
--- a/packages/core/src/Period.js
+++ b/packages/core/src/Period.js
@@ -76,7 +76,7 @@ export class Period extends TemporalAmount /* extends ChronoPeriod */ {
         const _months =  MathUtil.safeToInt(months);
         const _days = MathUtil.safeToInt(days);
 
-        if((_years || _months || _days) === 0){
+        if( _years === 0 && _months === 0 && _days === 0 ){
             if (!Period.ZERO) {
                 this._years = _years;
                 this._months =  _months;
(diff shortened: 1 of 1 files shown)


phueper:
THanks!  I would like to @pithu s oppinion on this ... The change is minimal and i think it works, but due to special JS Features...
A logical OR of (number) 0s results in (number) 0 ... and thus 0 === 0 is true ... but it only works due to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators ... the expr1 (0) does not evaluate to true, so expr2 (0) is the value of the expression...  that seems very implicit to me...

Somehow **i** would prefer a more explicit logic here (the proposed one makes my head spin ;)) ... something like 
``` 
if ((_year === 0) && (_month === 0) && (_days === 0)) {
...
```
i think this is way more readable and obvious ...  @pithu ? what do you think?

bowersj:
Thanks for your feedback.

Looking back, I realize I should have done what you proposed. I'll wait to hear from pithu before I make any changes though.

pithu:
> Somehow i would prefer a more explicit logic here (the proposed one makes my head spin ;)) ... something like
> `if ((_year === 0) && (_month === 0) && (_days === 0)) {`

agree

bowersj:
I'll make the change.

bowersj:
I finished those changes. Any more feed back? :)

pithu:
Lgtm

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #382, opened 2020-02-20
typescript typings for Duration ofSeconds


jseter:
Documentation and code represent this static method as:

public static | ofSeconds(seconds: Number, nanoAdjustment: Number): Duration

however typescript typing file is:
static ofSeconds(seconds: number): Duration;
instead of:
static ofSeconds(seconds: number, nanoAdjustment?: number): Duration;

pithu:
Hello @jseter 

thanks for reporting. The typescript definitions are community driven. We are open for pull requests.

PR #383, opened 2020-02-24
js-joda/js-joda#382 

jseter:
Added missing nanoAdjustment to Duration.ofSeconds static method.


pithu:
Thanks a lot great thing!

can you please add a test for it here https://github.com/js-joda/js-joda/blob/master/packages/core/test/typescript_definitions/js-joda-tests.ts#L516

PR #388, opened 2020-03-10
Fixed types of DateTimeFormatterBuilder.prototype.toFormatter to allow zero arguments

InExtremaRes:
Previously the parameter `resolverStyle` was required.


pithu:
Is there a test for this case ?

InExtremaRes:
I didn't add any with zero arguments but there was some that didn't break with this change. Do you want me to add an specific test for this case?

pithu:
Yes would be great to cover one case with and one without argument.

InExtremaRes:
Sorry for the delay. There were test cases with one argument and I just pushed a commit testing the types without any.

pithu:
@InExtremaRes thx a lot for fixing this

Issue #386, opened 2020-03-02
TypeScript declaration types for locales


InExtremaRes:
I started working in TS declarations for the locale package but have some questions:

1. Where should I put the file? In core and others the `.d.ts` is in the `dist` folder, but `locale` has many prebuilt packages and every one of them should have an `index.d.ts`. I guess in your infrastructure is possible to copy the same file to every locale specific package at build or package time, however this leads me to the next question.
2. Should every locale specific package have types for its own constants? I mean, if I import `'@js-joda/locale_en-us'` then I have the constants `Locale.US`, `Locale.ENGLISH`, but `Locale.FRENCH` will fail at runtime. Do you think is worthwhile to have independents types declarations for just the constants in every locale specific package?
3. I think the method `DateTimeFormatter.prototype.withLocale(locale)` should not be typed in core and be available only from the moment one imports one locale. This can be done with module augmentation in TypeScript. Is it OK for you? I think that would be a breaking change from the types perspective even though `withLocale` is useless without a proper `Locale` instance. If we can go that path then I suggest to delete the `Locale` interface in core, that is just a hack and allows to pass anything as a Locale (since `Locale` is an empty interface and in TS almost everything is assignable to `{}`).

Regards and good day.

phueper:
Hi @InExtremaRes,

thanks for working on this. My thoughts:

1. I would propose to first of all create a PR that puts a valid .d.ts file in packages/locale/dist . the prebuilt packages are all auto generated, so it doesn't make much sense putting anything in their dist folder... for the other files we use something like templates, or replace values from other files... if you could provide the diff you would imagine from the main d.ts to the locale specific d.ts we can maybe work it out... but as i said, first step would be the file in packages/locale/dist i think

2. i'm not sure wether thats possible... i think we are currently filling an availableLocales field when building the locale specific packages, but i'm not sure wether it is easily possible for the predefines ... for now i would just put all the predefined Locale in the .d.ts (they are available in JS anyway)

3. REgarding this i'm really unsure ... i would also like to get @pithu  s oppinion wether we want to split the .d.ts like this ... i think best would be to have a separate PR for it and discuss it there, i'm not a typescript expert, maybe the implications will become more obvious in a separate PR 🤔

pithu:
I would prefer to keep things simple. What typescript declarations should go to the locale package ? Can't we put these into core ?

InExtremaRes:
Hi @phueper, @pithu.

[...]

PR #389, opened 2020-03-10
New TS types for locale package

InExtremaRes:
As discussed in #386, this is my proposal.

The relevant points:

1. The `Locale` type is no longer in core since has no use without a locale imported.
2. The locale module add a `Locale` class that can be imported.
3. When the locale module is imported the module `'@js-joda/core'` is augmented, adding the `withLocale` method to `DateTimeFormatter` and other locale specific methods.

As you can see in the test, it just works. If you comment the line `import { Locale } from '../..';` in the test, then TS fails since the methods `withLocale`, `appendText`, etc. are not known.

I have no much experience with packaging and publishing so maybe there is more work needed.

I added types for `DateTimeFormatterBuilder.prototype.appendWeekField()` but couldn't find test for it. Is this method part of the public API? Are the types correct?


InExtremaRes:
Hi there. Any opinions on this? I hope you guys are ok.

InExtremaRes:
This PR has become stale. I can reopen it, it's a small PR, but I never get your thoughts so I don't know if it's worth it.

@phueper @pithu 

As I mentioned, TS types for locale-aware methods are very misleading since an empty interface is a wildcard target, almost everything can be assigned to it, and so it allows me to do this without complaining:

```js
DateTimeFormatter.ISO_INSTANT.withLocale([1, 2, 3]);
```

I'd love to have this included in the major release.

pithu:
> I'd love to have this included in the major release.

We can create another major, no worries 

@phueper any thoughts on this ?

phueper:
Hi,

First of all... sry for letting this go stale ... 

I do like it, i wonder wether removing the Locale from core is a problem but i guess it shouldn't be ... i think it makes sense to move it to the locale package(s) ... 

That said, i'm not sure about removing the `getDisplayName` i see that it includes a `Locale` parameter, but i don't know if it makes sense to remove it just for that...  couldn't we make the param an `any` or something similar? :thinking:

InExtremaRes:
> That said, i'm not sure about removing the `getDisplayName` i see that it includes a `Locale` parameter, but i don't know if it makes sense to remove it just for that... couldn't we make the param an `any` or something similar?

@phueper 

Maybe I'm misunderstanding something but I think those methods are just not implemented. For me they always fail with `IllegalArgumentException: Pattern using (localized) text not implemented yet!`. The documentation is not very clear here. How should those methods be used?

phueper:
@InExtremaRes 

they are implemented, but they only work if you have loaded @js-joda/locale... i think what they should output is the (localized) text for the type, so e.g. "Monday" or "Montag" (en-US and de-DE) for DayOfWeek, "January" and "Januar" accordingly for Month.... they are using functions in the formatter that are only available when @js-joda/locale is available and thus fail without it (i think, anyway :)... that's what i think should happen, otherwise i'd consider it a bug)

I'm not sure wether TS allows to "inject" methods into an already existing type, so could we only add them to the classes if @js-joda/locale is loaded? I mean since they could be used, i'm not sure wether it is correct to just drop them :thinking:

phueper:
sorry... i just realized, that the "inject" is already done in this PR for `withLocale` and others... so could we then add the `displayName` methods back also?

phueper:
*and* i just realized, that the methods actually are not implemented:

```
// eslint-disable-next-line no-unused-vars
    displayName(style, locale) {
        throw new IllegalArgumentException('Pattern using (localized) text not implemented yet!');
        // return new DateTimeFormatterBuilder().appendText(ChronoField.DAY_OF_WEEK, style).toFormatter(locale).format(this);
    }
```

so... i guess this should be a bug

InExtremaRes:
@phueper 

No problem and thanks for clarifying. The proposal in this PR is exactly that: allowing locale TS declarations to inject (or _augment_, in TS terms) the types into the classes of the core module, just mirroring the JS implementation. Since I couldn't find implementations for those methods I just removed their declarations.

I still think the documentation is not very clear, though And requires a good update: [here](https://js-joda.github.io/js-joda/manual/formatting.html#formatting-with-locales) for instance still references the old packages, before de mono-repo. The [README](https://github.com/js-joda/js-joda/blob/056ddca6f6271e0f0366e416e96832d31d4d4c1b/packages/locale/README.md) is updated but lacks a good explanation of what exactly _changes_ when the module is imported.

@phueper @pithu 

I will rebase and force push this PR (if you are ok with that) to update all that have changed recently. If those implementations are missing right now I'd open an issue and just remove them from TS declarations, adding them back if they got to be implemented; unless you have plans to implement them in the short term.

Another question: The locale module defines a class [WeekFields](https://github.com/js-joda/js-joda/blob/89bbbaa967532a8b4cdc3d65407ad5fa93a91e06/packages/locale/src/temporal/WeekFields.js#L552) but honestly I don't understand if that is meant to be public, importable somehow, or is just an implementation detail. Can be used by user code and TS declarations should be made for it as well?

Many thanks to both of you.

phueper:
@InExtremaRes 

i'm fine with rebasing/force pushing the PR and removing the displayName definitions and opening an issue

Regarding WeekFields: in threeten it is a public class https://www.threeten.org/threetenbp/apidocs/org/threeten/bp/temporal/WeekFields.html so i think it is part of the API, it stores the dayOfWeek/weekOfMonth/weekOfYear definition for different locales (e.g. wether a week starts with sunday or monday, ...) so i think it *could* be useful even though i can't think of a concrete usecase of the top of my head, but i think it should be in the TS definitions as part of the API

InExtremaRes:
This is ready to be re-reviewed. I removed the methods and opened #424. I end up not typing `WeekField` because (I think) it is not publicly available in the bundled js, neither in core nor in locale, don't know if that is also a bug.

packages/locale/test/typescript_definitions/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+	"compilerOptions": {
+		"target": "ES6",
+		"lib": ["ES2015"],
+		"module": "CommonJS",
+		"pretty": true,
```
phueper, line 6: If pretty is true here, do we need the `--pretty` in the command in package.json?

InExtremaRes: Not at all! I just missed it. I'll fix ASAP.

InExtremaRes:
@pithu @phueper Hi there! Could you take another look at this? Last comment was addressed. Thanks.

phueper:
again, sorry for the long time... i think we can and should merge this

open points that come to my mind:

[ ] we need to look at copying the .d.ts files to the locale prebuilt packages i think so all the @js-joda/locale_xx packages get the ts definitions as well
[ ] we need to decide what we do with version, i would guess it is an API breaking change in core as well as in locale :thinking:

InExtremaRes:
> [ ] we need to look at copying the .d.ts files to the locale prebuilt packages i think so all the @js-joda/locale_xx packages get the ts definitions as well

I think you're right, every package should have it's own .d.ts. Sadly, as I said, I don't have much experience with packaging and publishing. Any advice here is very welcome.

> [ ] we need to decide what we do with version, i would guess it is an API breaking change in core as well as in locale

Most users won't notice anything with the change, but it is technically breaking from the types perspective of both packages, so I think is best to treat it as such.

phueper:
> I think you're right, every package should have it's own .d.ts. Sadly, as I said, I don't have much experience with packaging and publishing. Any advice here is very welcome.
> 
i'll try to come up with a solution, shouldn't be too hard ...

phueper:
@InExtremaRes  i added the needed changes for the generated packages (i think) to https://github.com/phueper/js-joda/tree/InExtreamRes-locale-types

I don't want to push into your repo (and honestly i don't even now if i can) thus i branched yours and added the necessary commits... could you merge them into your branch and test wether  the TS typings then work for one of the local @js-joda/locale_xx packages ? If that works, i think we can then merge the PR.. incrementing the major versions i'll do in master afterwards i think when we are ready to publish

InExtremaRes:
@phueper Thank you!

> could you merge them into your branch and test wether the TS typings then work for one of the local @js-joda/locale_xx packages ?

Do you mean test it locally or make an automated test for it?

phueper:
> Do you mean test it locally or make an automated test for it?
just test it locally, if the typings are recognized for a locale specific pacakge ... just to make sure, that the changes i made work :) that would be great

InExtremaRes:
It seems to work perfectly.

I created a new npm project, installed `typescript`, `ts-node` (to run without transpile first) and manually copied the js-joda files in the corresponding node_modules folders. In particular I copied the content of `locale/packages/es` into `node_modules/@js-joda/locale_es` (I did something similar for `core`).

This GIF shows the use inside the editor and the result of the code:
js-joda-locale-es

As you can see, the `DateTimeFormatter` doesn't have the `withLocale()` method until I import the package.

As I side note, I have some pending changes that could fit into a new major release. I will try to make a PR today or tomorrow.

phueper:
nice ! Thanks again for this, i'll wait with merging until we are ready for the major version update

InExtremaRes:
@phueper I have a little problem. I'm starting to need the previous PRs (like this one) in the next changes I'm working on, mainly to avoid conflicts, since I'm touching the sames files. I can make a bigger PR where each commit is a single unit, but even if the changes are related they can be accepted or rejected separately so not a great choice I think. I also can merge these branches into the new one, but the history will become complex after a few branches (I'm fine with that, though).

What do you suggest? Maybe this PR and #456 can be merged, even in master or another branch I can take as base for the next ones.

PR #401, opened 2020-06-02
Github actions test

pithu:


pithu:
Experience with github actions was not that great, travis is doing a good job, close this for now

PR #404, opened 2020-06-05
Some fixes to TS declarations

InExtremaRes:
- Some `FROM` static properties were correctly typed as `TemporalQuery` but for the wrong `Temporal`.
- `Duration` were missing methods `minus` and `plus` with a single `Duration`. The old methods were also wrong since `plus` previously accepted two arguments of type `number`.
- `ZonedDateTime` were missing methods `withEarlierOffsetAtOverlap` and `withLaterOffsetAtOverlap`.

All of them were fixed and tests added. Also I left some TODO comments for methods that are meant to be internal only, like `minusDuration` or `plusTemporalAmount`. This should be removed in the next major version since TS user could be relying on them. Or, if you want, I can remove them in this PR.

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -117,9 +117,12 @@ export class Duration extends TemporalAmount {
     get(unit: TemporalUnit): number;
     isNegative(): boolean;
     isZero(): boolean;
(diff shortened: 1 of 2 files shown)


pithu:
This sounds like a never ending story and i thought everything is cleaned up already. Thanks a lot for your effort! I would leave the major change for later, but its good to have the comments.

PR #407, opened 2020-06-14
Error base name changed to "JsJodaException" and exceptions exposed in TS

InExtremaRes:
The name of the function that is base of the rest of errors was previously `E`. The problem is that the name of the constructor function is usually exposed when inspecting through the console, and `E` is not very very descriptive. There are some tricks to compute dynamically the name of a function based on a given value, but those doesn't play well in ES5 environments. Instead, I changed the name of the function to `JsJodaException` that is more descriptive of where the error comes from if it is just console.logged.

I also added type declarations for exceptions classes as they can be used for user to check the name or via `instanceof`.

### DIFF
--- a/packages/core/src/errors.js
+++ b/packages/core/src/errors.js
@@ -4,8 +4,8 @@
  */
 
 function createErrorType(name, init, superErrorClass = Error) {
-    function E(message) {
-        if (!Error.captureStackTrace){
+    function JsJodaException(message) {
+        if (!Error.captureStackTrace) {
             this.stack = (new Error()).stack;
         } else {
             Error.captureStackTrace(this, this.constructor);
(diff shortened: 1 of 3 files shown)


packages/core/src/errors.js
```diff
@@ -16,10 +16,10 @@ function createErrorType(name, init, superErrorClass = Error) {
             return `${this.name}: ${this.message}`;
         };
     }
-    E.prototype = new superErrorClass();
-    E.prototype.name = name;
-    E.prototype.constructor = E;
-    return E;
+    JsJodaException.prototype = Object.create(superErrorClass.prototype);
```
InExtremaRes, line 19: Previously a full instance of the superClass was created just to inherit from it, but a simple object with the right prototype is enough.

InExtremaRes:
I didn't add any since most of those classes do nothing more than exist. But I can add some to test the names can be imported and are, in fact, `Error`s.

pithu:
Thanks a lot @InExtremaRes for your effort and for the nice tests!

PR #408, opened 2020-06-20
Fixes and cleanups to TemporalFields and TemporalUnits

InExtremaRes:
This PR makes some cleanups and improvements to temporal fields and units. Some of them were mentioned in #355. In particular:

- The "abstract" class `TemporalField` now contains the relevant methods, all documented.
- `ChronoField` now have methods `isSupportedBy` and `adjustInto` to conform the `TemporalField` interface. The implementation is the same as in threetenbp. I made some basic tests, mostly to prove that the methods exist and work with a particular example; the core of the implementation is already tested in the temporal objects.
- Some methods still named as `getDisplayName` were renamed to `displayName`.
- `IsoFields` is no longer a class, but is now and object to better serve its purpose as an _enum_. The properties are still late-setted in the `_init()` function.
- The ThreeTen indicates that `checkValidValue` in a `ChronoField` should return the given value (or throw), but `ValueRange.prototype.checkValidValue` returned `undefined`. The mentioned function now returns the given value. I tested this from `ChronoField.prototype.checkValidValue` since its more important than `ValueRange` itself.
- The TS declarations were properly updated.
- Since I were already into that, I updated the documentation of `TemporalField`, `ChronoField`, `IsoFields`, `TemporalUnit` and `ChronoUnit`. Some methods had an incorrect signature in the jsdoc. I also added the descriptions for the fields itself.
    I changed the docs for the fields into a list. I think it looks better on the generated page. I also removed the docs in the functions that are inherited from a base class; esdoc copies the documentation, but sadly not the signature, so the signature need to stay.
    Also added some documentation to the TS declarations for the methods and classes I changed in this PR. The documentation is mostly a brief description, but I think it can help developers using an editor/IDE with TS inference. Probably in the future I can improve this a little further.

### DIFF
--- a/packages/core/src/temporal/ChronoField.js
+++ b/packages/core/src/temporal/ChronoField.js
@@ -22,68 +22,189 @@ import {YearConstants} from '../YearConstants';
  * just with slightly different rules.
  * The documentation of each field explains how it operates.
  *
- * ### Static properties of Class {@link ChronoField}
+ * ### Static properties:
  *
- * ChronoField.NANO_OF_SECOND
- *
- * ChronoField.NANO_OF_DAY
- *
- * ChronoField.MICRO_OF_SECOND
- *
- * ChronoField.MICRO_OF_DAY
- *
- * ChronoField.MILLI_OF_SECOND
- *
- * ChronoField.MILLI_OF_DAY
- *
- * ChronoField.SECOND_OF_MINUTE
- *
- * ChronoField.SECOND_OF_DAY
- *
- * ChronoField.MINUTE_OF_HOUR
- *
- * ChronoField.MINUTE_OF_DAY
- *
- * ChronoField.HOUR_OF_AMPM
- *
- * ChronoField.CLOCK_HOUR_OF_AMPM
- *
- * ChronoField.HOUR_OF_DAY
- *
- * ChronoField.CLOCK_HOUR_OF_DAY
- *
- * ChronoField.AMPM_OF_DAY
- *
- * ChronoField.DAY_OF_WEEK
- *
- * ChronoField.ALIGNED_DAY_OF_WEEK_IN_MONTH
- *
- * ChronoField.ALIGNED_DAY_OF_WEEK_IN_YEAR
- *
- * ChronoField.DAY_OF_MONTH
- *
- * ChronoField.DAY_OF_YEAR
- *
- * ChronoField.EPOCH_DAY
- *
- * ChronoField.ALIGNED_WEEK_OF_MONTH
- *
- * ChronoField.ALIGNED_WEEK_OF_YEAR
- *
- * ChronoField.MONTH_OF_YEAR
- *
- * ChronoField.PROLEPTIC_MONTH
- *
- * ChronoField.YEAR_OF_ERA
- *
- * ChronoField.YEAR
- *
- * ChronoField.ERA
- *
- * ChronoField.INSTANT_SECONDS
- *
- * ChronoField.OFFSET_SECONDS
- *
+ * - `ChronoField.ALIGNED_DAY_OF_WEEK_IN_MONTH`: This represents concept of the count of
+ * days within the period of a week where the weeks are aligned to the start of the month.
+ * This field is typically used with `ALIGNED_WEEK_OF_MONTH`.
+ * 
+ * - `ChronoField.ALIGNED_DAY_OF_WEEK_IN_YEAR`: This represents concept of the count of days
+ * within the period of a week where the weeks are aligned to the start of the year.
+ * This field is typically used with `ALIGNED_WEEK_OF_YEAR`.
+ * 
+ * - `ChronoField.ALIGNED_WEEK_OF_MONTH`: This represents concept of the count of weeks within
+ * the period of a month where the weeks are aligned to the start of the month. This field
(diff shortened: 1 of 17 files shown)


InExtremaRes:
I notices the coverage decreased a little. I guess it's because of the new methods on `TemporalField`. How can I get a report of the lines and branches uncovered?

pithu:
> I notice the coverage decreased a little. I guess it's because of the new methods on `TemporalField`. How can I get a report of the lines and branches uncovered?

https://coveralls.io/builds/31581070

InExtremaRes:
Thanks. And yes, the uncovered lines are the new abstract methods on `TemporalField`. Do you think it's worth testing that those methods fail? I created those methods mostly for consistency.

pithu:
I could take just a short look now. Just on thing i stumbled over, it would be helpful if you take major changes out of the PR and discuss it in another PR. eq renaming `getDisplayName` to `displayname`.  

I hope, i will find some more time next week

phueper:
> Thanks. And yes, the uncovered lines are the new abstract methods on `TemporalField`. Do you think it's worth testing that those methods fail? I created those methods mostly for consistency.

imho we should create tests that make sense for the functionality of the library.  In this case i'm not really sure... i don't think the tests would help understand the functionality, but just increase coverage, so i guess we could leave them as it is and accept the decrease in coverage.
I'm not sure wether we do test other "abstract" functions, but i would think most of the tests really are about the functionality.
That said, if we make the decision to not test these lines maybe you could add some `/* istanbul ignore` lines and comment to make it clear why there are no tests for this file ?

InExtremaRes:
I totally agree with @phueper, I think to increase coverage just for the sake of it is not worth it, but I prefer to ask since some teams find value on that.

If you think is better to split this PR I'll send some new ones in the next days and then close this.

packages/core/src/temporal/ChronoField.js
```diff
@@ -120,58 +242,73 @@ export class ChronoField extends TemporalField {
...
+     *
+     * This method checks against the range of the field in the ISO-8601 calendar system.
      *
-     * @param {number} value
-     * @returns {*}
+     * @param {!number} value the value to check.
+     * @returns {number} the value that was passed in.
      */
     checkValidValue(value) {
         return this.range().checkValidValue(value, this.name());
```
pithu, line 291: Looks like, i found an existing bug, should be `...checkValidValue(value, this)`. Doesn;t harm because the field is just used for printing

packages/core/src/temporal/ChronoField.js
```diff
@@ -217,83 +352,57 @@ export class ChronoField extends TemporalField {
...
-     * It is recommended to use the second approach, {@link range},
-     * as it is a lot clearer to read in code.
-     *
-     * Implementations should perform any queries or calculations using the fields
-     * available in {@link ChronoField}.
-     * If the field is not supported a {@link DateTimeException} must be thrown.
-     *
-     * @param {!TemporalAccessor} temporal - the temporal object used to refine the result, not null
-     * @return {ValueRange} the range of valid values for this field, not null
-     * @throws DateTimeException if the range for the field cannot be obtained
```
pithu, line 247: Did you moved this documentation to somewhere or just deleted it ?

packages/core/src/temporal/TemporalField.js
```diff
@@ -21,4 +23,226 @@
...
+     *   // these two lines are equivalent, but the second approach is recommended
+     *   temporal = thisField.rangeRefinedBy(temporal);
+     *   temporal = temporal.range(thisField);
+     * </pre>
+     * It is recommended to use the second approach, {@link range},
+     * as it is a lot clearer to read in code.
+     *
+     * Implementations should perform any queries or calculations using the fields
+     * available in {@link ChronoField}.
+     * If the field is not supported a {@link DateTimeException} must be thrown.
```
pithu, line 115: Ah here it is ;)

pithu:
Does anyone knows why the status reports from travis are not returning ? The corresponding tests in travis arre green.

packages/core/src/temporal/ChronoField.js
InExtremaRes: I'll take a look into this.

InExtremaRes:
> Amazing piece of work, well done! A good improvement of documentation and cleanup. The generated esdocs looks much better.

Glad to hear that :)

> Beside the major changes, i would merge it, then make a minor release first. Do you have more major changes in your pipeline ?

Yes I do. I'd like to cleanup a little more, address some of the issues commented in #355, refine TS declarations and more. I'll be sending more PR the next days. I think makes sense a minor release from the current state (before this PR).

packages/core/src/temporal/ChronoField.js
InExtremaRes: Well, fixed. As you said, since it's just for string concatenation there is no difference. Better to be consistent, nonetheless.

pithu: Yes, it was definitely a bug

InExtremaRes:
If I update the branch, should travis run again?

pithu:
Worx for me. 🤞

InExtremaRes:
CI is green :tada: When you merge I'll send another PR based on this.

pithu:
well done, thanks a lot, next update will be a major!

PR #415, opened 2020-06-26
TS types refactor part 1

### LINKED ISSUE #381: Issue with Typescript/Flow types : minusTemporalAmount & plusTemporalAmount
The type definitions still contain minusTemporalAmount & plusTemporalAmount functions but they seem to have been removed as part of the migration into the js-joda npm org

InExtremaRes:
This is part 1 of a series of improvements and refactors to TS declarations.

From an user perspective I think the most important changes are:

- Methods of `TemporalAccesor` and `Temporal` was not properly marked as `abstract`. Now all known methods derived from this classes are typed consistently.
- `LocalTime.prototype.get()` took a `ChronoField` instead of a `TemporalField`. Same for `Year`.
- Is expected that a `TemporalQuery` returns `null` but `TemporalAccesor.prototype.query` was not marked as such.
- `LocalTime.prototype.until` now correctly allow only `Temporal` instances while previously accepted `TemporalAccesor`.

Some internal methods were removed:
- `Duration.prototype.`
	- `minusAmountUnit`
	- `minusDuration`
	- `plusAmountUnit`
	- `plusDuration`
- `ZonedDateTime.prototype.plusTemporalAmount`.

I also added types for some methods, like `toJSON` and `equals` to enum-like classes such as `TextStyle`. While I'd like to remove `getLong`, its types were added since is the way js-joda works right now.

I sorted the methods of some classes to facilitate myself catching mistakes, that's why`YearMonth` has a greater diff. Additionally I changed the way `tsc` was invoked from CLI arguments to a `tsconfig.json`, this way is easier to see errors on supported IDEs.

Closes #381

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -85,6 +85,7 @@ export class DayOfWeek extends TemporalAccessor implements TemporalAdjuster {
     name(): string;
     ordinal(): number;
     plus(days: number): DayOfWeek;
+    toJSON(): string;
     toString(): string;
     value(): number;
 }
@@ -207,6 +208,10 @@ export class ResolverStyle {
     static LENIENT: ResolverStyle;
 
     private constructor();
+
+    equals(other: any): boolean;
+    toJSON(): string;
+    toString(): string;
 }
 
 export class SignStyle {
@@ -215,6 +220,12 @@ export class SignStyle {
     static ALWAYS: SignStyle;
     static EXCEEDS_PAD: SignStyle;
     static NOT_NEGATIVE: SignStyle;
+
+    private constructor();
+
+    equals(other: any): boolean;
+    toJSON(): string;
+    toString(): string;
 }
 
 export class DateTimeFormatter {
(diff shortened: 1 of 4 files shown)


InExtremaRes:
While working on this I saw again a `LocalDate.EPOCH_0` constant. Is it a public property? The ThreeTen doesn't have that and the name looks bit weird.

pithu:
> While working on this I saw again a `LocalDate.EPOCH_0` constant. Is it a public property? The ThreeTen doesn't have that and the name looks bit weird.

Might be another "creative extension", probably we should remove this as well.

InExtremaRes:
> Might be another "creative extension", probably we should remove this as well.

There're places where js-joda deviates slightly from ThreeTen, adding some conveniences (`isoWeekyear()` as en example). I think there is no harm on keeping a constant for an epoch date, but then the constant should be renmed to `EPOCH` to be consistent.

In any case, I'd consider that for another PR since there is also JS code to adjust, keeping this TS specific.

OTOH, is travis acting weird again?

PR #416, opened 2020-06-29
Implement OffsetTime and OffsetDateTime

### LINKED ISSUE #240: OffsetTime missing

exoego:
Closes #240

Thanks for creating this repository.
Recently, I have an increased need for `OffsetTime` and `OffsetDateTime`, so I am trying to add them.

I used the below source files as reference implementation.

* src
  * https://github.com/ThreeTen/threetenbp/blob/800401d216666575853412a459f8e8a83af95709/src/main/java/org/threeten/bp/OffsetDateTime.java
  * https://github.com/ThreeTen/threetenbp/blob/800401d216666575853412a459f8e8a83af95709/src/main/java/org/threeten/bp/OffsetTime.java
* test
  * https://github.com/ThreeTen/threetenbp/blob/800401d216666575853412a459f8e8a83af95709/src/test/java/org/threeten/bp/TestOffsetDateTime.java
  * https://github.com/ThreeTen/threetenbp/blob/800401d216666575853412a459f8e8a83af95709/src/test/java/org/threeten/bp/TestOffsetTime.java

IIUC, their license (BSD-3) is compatible with this repository. 

Please be aware that some extensive test cases are really slow, so I added `this.timeout(50000);` to extend timeout.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "prepublish": "npm run build-dist",
-    "test": "NODE_ENV=test ./node_modules/.bin/mocha --timeout 5000 --require @babel/register ./test/*Test.js ./test/**/*Test.js ./test/**/**/*Test.js ./test/*Test_mochaOnly.js",
+    "test": "NODE_ENV=test ./node_modules/.bin/mocha --timeout 5000 --require @babel/register ./test/*Test_mochaOnly.js ./test/*Test.js ./test/**/*Test.js ./test/**/**/*Test.js",
     "test-coverage": "NODE_ENV=test COVERAGE=1 ../../node_modules/.bin/nyc --report-dir=build/coverage --reporter=lcov --reporter html ./node_modules/.bin/mocha --timeout 5000 --require @babel/register --reporter progress ./test/*Test.js ./test/**/*Test.js ./test/**/**/*Test.js",
     "test-browser": "./node_modules/.bin/karma start --reporters=dots --single-run",
     "test-saucelabs": "./node_modules/.bin/karma start --reporters=\"dots,saucelabs\" --browsers=\"sl_chrome,sl_ie,sl_firefox,sl_ios_simulator\" --single-run=true",
--- a/packages/core/src/Instant.js
+++ b/packages/core/src/Instant.js
@@ -11,6 +11,7 @@ import {Clock} from './Clock';
 import {LocalTime} from './LocalTime';
 import {ZonedDateTime} from './ZonedDateTime';
 import {MathUtil} from './MathUtil';
+import {OffsetDateTime} from './OffsetDateTime';
 
 import {Temporal} from './temporal/Temporal';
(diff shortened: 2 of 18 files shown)


exoego:
Only the below 2 test cases (which seems unrelated to this change) are failing... any help 🤔 ?

```
@js-joda/core:   1) plugTest
@js-joda/core:        should not fail using the localDateConvenientPlugin:
@js-joda/core:      TypeError: Cannot read property 'prototype' of undefined
@js-joda/core:       at fn (test/plugTest_mochaOnly.js:17:26)
@js-joda/core:       at use (src/use.js:15:13)
@js-joda/core:       at Context.<anonymous> (test/plugTest_mochaOnly.js:26:9)
@js-joda/core:       at processImmediate (internal/timers.js:456:21)
@js-joda/core:   2) plugTest
@js-joda/core:        should use the previously added plugin:
@js-joda/core:      TypeError: date1_.isAfterOrEqual is not a function
@js-joda/core:       at Context.<anonymous> (test/plugTest_mochaOnly.js:38:23)
@js-joda/core:       at processImmediate (internal/timers.js:456:21)
```

pithu:
Hi @exoego 

Thanks a lot for the amazing work, all in all it looks good to me.

The test failure is probably one of this annoying cycle dependency issues. I could fix it by running the plugin test first. You can give the following change in the package.json a try:
```
   ...
    "test": "NODE_ENV=test ./node_modules/.bin/mocha --timeout 5000 --require @babel/register ./test/*Test_mochaOnly.js ./test/*Test.js ./test/**/*Test.js ./test/**/**/*Test.js",
   ...
```

The extended test timeouts are a problem. As far as i remember we had a lot of issues with flaky travis builds, because of that. We solved that by not iterating over all cases as the the ThreeTen tests are doing. Not the best solution, but i don't have a better fix and i doubt that tests in that detail bring any value. 
We also added some "reduced" tests for the browser and test coverage runners, you can find examples by searching for the usage of `isBrowserTestRunner` and `isCoverageTestRunner`.
Would be good if you could reduce test time, currently i guess most test are running below 200ms on travis.

exoego:
@pithu Thanks for guidance.
Tests are reordered and slow tests are reduced.

pithu:
Cool, that's fast ! 

Could you try to increase test coverage? Coverage decrease with your PR and looking into coverage reports it looks like there are reasonable candidates for tests https://coveralls.io/builds/31846690, eg tests for methods `until`, `isSupported`, `atDate`. 
And it would be also nice to have coverage for the new functionality in `Instant` and `LocalDate`.

exoego:
OK, I will add more tests to increase coverage.

exoego:
Now, newly added classes are 100% branch covered !!

pithu:
Great contribution, thanks a lot!

PR #417, opened 2020-06-30
Duration#(minus|plust) should accept TemporalUnit, not just ChronoUnit

exoego:
According to ThreeTen reference implementation and OracleJDK Javadoc, `Duration#minus` and `Duration#plus` should accept `TemporalUnit`, which is a super type of `ChronoUnit`.

https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#plus-long-java.time.temporal.TemporalUnit-
https://github.com/ThreeTen/threetenbp/blob/800401d216666575853412a459f8e8a83af95709/src/main/java/org/threeten/bp/Duration.java#L635

https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#minus-long-java.time.temporal.TemporalUnit-
https://github.com/ThreeTen/threetenbp/blob/800401d216666575853412a459f8e8a83af95709/src/main/java/org/threeten/bp/Duration.java#L793

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -118,7 +118,7 @@ export class Duration extends TemporalAmount {
     isNegative(): boolean;
     isZero(): boolean;
     minus(duration: Duration): Duration;
-    minus(amount: number, unit: ChronoUnit): Duration;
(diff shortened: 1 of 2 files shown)


pithu:
thx a lot for fixing that

PR #420, opened 2020-07-04
Cache Cldr operations to improve performance when printing timezones

### LINKED ISSUE #394: Performance degraded when printing timezones
Love js-joda, such a great way to work with dates and times in JS, but noticed today some very laggy performance in a table I had where I was using `js-joda` to display some datetimes in a `dd-MMM-yyyy h:MM a z` format. Delving in to the performance issues, the biggest issues were coming from locales calling `Cldr.load` every time I printed a datetime with a timezone name.  See below for full details, but adding in a timezone name makes things around the magnitude of 100x slower:

**Benchmark Code:**
```js
(global as any).testSpeed = () => {
  const NUM_TIMES = 1000;

  const formatterWithoutTimezone = DateTimeFormatter.ofPattern('dd-MMM-yyyy h:MM a').withLocale(Locale.US);
  const formatterWithTimezone = DateTimeFormatter.ofPattern('dd-MMM-yyyy h:MM a z').withLocale(Locale.US);

  const utcDateTime = ZonedDateTime.of(LocalDateTime.of(2020, 1, 2, 4, 5, 6), ZoneId.UTC);
  const nycDateTime = ZonedDateTime.of(LocalDateTime.of(2020, 1, 2, 4, 5, 6), ZoneId.of('America/New_York'));

  console.time('UTC without timezone');
  for (let i = 0; i < NUM_TIMES; ++i) {
    utcDateTime.format(formatterWithoutTimezone);
  }
  console.timeEnd('UTC without timezone');

  console.time('UTC with timezone');
  for (let i = 0; i < NUM_TIMES; ++i) {
    utcDateTime.format(formatterWithTimezone);
  }
  console.timeEnd('UTC with timezone');

  console.time('NYC without timezone');
  for (let i = 0; i < NUM_TIMES; ++i) {
    nycDateTime.format(formatterWithoutTimezone);
  }
  console.timeEnd('NYC without timezone');

  console.time('NYC with timezone');
  for (let i = 0; i < NUM_TIMES; ++i) {
    nycDateTime.format(formatterWithTimezone);
  }
  console.timeEnd('NYC with timezone');
}
```

**Results**:
```
UTC without timezone: 19.212158203125ms (probably just slower because of a cache warming thing being first)
UTC with timezone: 11.41796875ms
NYC without timezone: 5.911865234375ms
NYC with timezone: 599.635009765625ms
```

**Analysis**:
Luckily, it's pretty easy to profile and see where the time is being eaten up, and it's in the locale's `CldrZoneTextPrinterParser.js` `print` function (bottom-up view):

image

Specifically, the call to `Cldr.load` is the biggest culprit:

https://github.com/js-joda/js-joda/blob/ff6785aa3d13f637544e78c59334662833c4d861/packages/locale/src/format/cldr/CldrZoneTextPrinterParser.js#L108

**Solution?**
Could this call be cached somehow so that it only has to happen the first time something is printed? This same perf issue may be happening in other situations also, as I do see the same calls being made in a locale's in the `CldrZoneTextPrinterParser.js`  `parse` function also.

JKillian:
Fixes #394 by adding module-level caches and memoizing functions for various operations involved in printing timezone names.

Running the same test as in #394 that prints timezone names 1000 times, we now have a drastic speedup. 

**Before**:
```
UTC without timezone: 19.212158203125ms
UTC with timezone: 11.41796875ms
NYC without timezone: 5.911865234375ms
NYC with timezone: 599.635009765625ms
```

**After**:
```
UTC without timezone: 20.73486328125ms
UTC with timezone: 10.364990234375ms
NYC without timezone: 6.211181640625ms
NYC with timezone: 6.883056640625ms
```

My main question is if it's okay to have essentially global, permanent, caches as are added in this PR? Obviously there could be memory usage concerns with these constructs. I've left a few notes inline to summarize my thoughts.


packages/locale/src/format/cldr/CldrCache.js
```diff
@@ -0,0 +1,54 @@
+/*
+* @copyright (c) 2020, Philipp Thuerwaechter & Pattrick Hueper
+* @license BSD-3-Clause (see LICENSE.md in the root directory of this source tree)
+*/
+
+import cldrData from 'cldr-data';
+import Cldr from 'cldrjs';
+
+const cldrDataLoaded = new Set();
```
JKillian, line 9: This 'cache' seems harmless to me, it's just a set of strings passed to `cldrData()` so I imagine the size should be pretty trivial.

packages/locale/src/format/cldr/CldrCache.js
```diff
@@ -0,0 +1,54 @@
...
+* Loads the Cldr data for the given path if it hasn't been loaded before.
+*/
+export const loadCldrData = (path) => {
+    if (!cldrDataLoaded.has(path)) {
+        Cldr.load(cldrData(path));
+        cldrDataLoaded.add(path);
+    }
+};
+
+const localeToCldrInstanceCache = {};
```
JKillian, line 20: I _think_ each Cldr instance is fairly small memory-wise (since I imagine it just holds a few references to the already loaded Cldr data and doesn't copy it or anything) so this cache seems fairly harmless memory-wise to me.

Cldr instances do [have some sort of per-instance event listener system](https://github.com/rxaviers/cldrjs#cldreventjs), but I don't think that's being used anywhere, so sharing a single instance per locale string should be fine runtime-behavior-wise.

packages/locale/src/format/cldr/CldrCache.js
```diff
@@ -0,0 +1,54 @@
...
+* Memoized, so a given locale will always return the exact same cldr instance.
+*/
+export const getOrCreateCldrInstance = (locale) => {
+    if (localeToCldrInstanceCache[locale] == null) {
+        localeToCldrInstanceCache[locale] = new Cldr(locale);
+    }
+    return localeToCldrInstanceCache[locale];
+};
+
+const cldrInstanceToMapZonesCache = new Map();
```
JKillian, line 32: This cache I imagine could be a bit bigger, since it's making a new `mapZones` object for each provided cldr instance. I'm not sure how often this even ends up getting used when printing a timezone? So I could probably just get rid of this cache and most cases wouldn't see a major performance difference.

packages/locale/src/format/cldr/CldrZoneTextPrinterParser.js
```diff
@@ -25,6 +24,18 @@ const LENGTH_COMPARATOR = (str1, str2) => {
...
+/**
+ * Cache for `_cachedResolveZoneIdText`.
+ * 
+ * Its basic structure is:
+ * Map { cldrInstance: zoneId }
+ * Obj { zoneId: style}
+ * Obj { style: type}
+ * Obj { type: resolvedZoneIdText}
+ */
+const resolveZoneIdTextCache = new Map();
```
JKillian, line 37: Technically this cache could get really big if you have a large number of `cldr locales x zone ids to print x styles x types`, but that seems unlikely in most use cases. Without this cache, our 1000 iterations take about 90ms, with this cache, things take the 7ms in the PR summary.

packages/locale/src/format/cldr/CldrZoneTextPrinterParser.js
```diff
@@ -61,6 +95,7 @@ export default class CldrZoneTextPrinterParser {
                     if (metaZoneData) {
                         return metaZoneData;
                     } else {
+                        const mapZones = getOrCreateMapZones(cldr);
```
JKillian, line 98: Instead of computing and passing `mapZones` in everytime, it's now only computed if we get to the part of the function where it's actually needed.

packages/locale/src/format/cldr/CldrCache.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+* Returns a map zones object for a cldr instance.
+* Memoized, so a given cldr instance will always return the exact same object.
+*/
+export const getOrCreateMapZones = (cldr) => {
+    if (!cldrInstanceToMapZonesCache.has(cldr)) {
```
pithu, line 38: Would it possible to change the key from a cldr instance to a locale string ? It is very hard to read and understand why this is mapped to an instance. A locale string would make this much more clear IMO.

Regarding cache size, again, some numbers would be great.

packages/locale/src/format/cldr/CldrZoneTextPrinterParser.js
```diff
@@ -35,11 +46,34 @@ export default class CldrZoneTextPrinterParser {
         requireNonNull(textStyle, 'textStyle');
         requireInstance(textStyle, TextStyle, 'textStyle');
         this._textStyle = textStyle;
-        Cldr.load(cldrData('supplemental/likelySubtags.json'));
-        Cldr.load(cldrData('supplemental/metaZones.json'));
+        loadCldrData('supplemental/likelySubtags.json');
+        loadCldrData('supplemental/metaZones.json');
     }
 
-    _resolveZoneIdText(cldr, zoneId, style, type, mapZones) {
```
pithu, line 42: Further down, this method is stilled called with mapZones, they should probably replaced by calls to `_cachedResolveZoneIdText`

packages/locale/src/format/cldr/CldrCache.js
JKillian: > I wonder why you use an object here and for the other cache a new Map(), any reason for that ?

Not any great reason - I was just generally using a `Map` where forced to because of using objects as keys, and using a regular object everywhere else.

> About cache size, i am unsure, i don't know what the Cldr is storing internally. Could we somehow measure memory usage ? Like write a test with a typical scenarios and capture memory usage, so we could get an idea, how much this could get.

I'm not sure either how much extra memory each instance uses. I think doing a quick test to try to verify that makes sense, in case it is a lot of copied data per instance.

packages/locale/src/format/cldr/CldrCache.js
JKillian: Yes, should be easy to switch to make this take a locale instead! I'll do that

phueper:
I like the PR ... i never thought of the cldr (which is actually just loading/requiring the cldr data) to make this big a difference, but it does make sense.
I would also like to get some numbers on the memory usage if that is possible.

i wonder wether this might hurt us if we ever get to fix https://github.com/js-joda/js-joda-locale/issues/93  ... i.e. when loading multiple different locale instances ... but i can't think of a reason right now ... 

So, if the review comments by @pithu are resolved and we have some memory numbers for the tests... i'd be happy to get this merged :+1:

packages/locale/src/format/cldr/CldrCache.js
JKillian: I wrote a loop in some test initialization code that just created 100 Cldr instances, and then used the Chrome allocation recorder to measure what was allocated before and after that loop. It looks like the Cldr instances are fairly harmless, a little less than a KB each (see screenshot below), so it seems that keeping them around should be fine.

I'm not 100% confident that this was an accurate test, just because I'm not too familiar with the tools around measuring memory allocations, but as far as I can tell, each instance is pretty light.

image

packages/locale/src/format/cldr/CldrCache.js
JKillian: Each new `mapZones` objects appears to be about 20KB.  I'm not quite sure on the context of this code - is it likely that we generate different `mapZones` objects for many different locales over the lifetime of a program? Or is it typically only a few locales? If the number is small enough, probably fine to keep this cache around

image

JKillian:
I just pushed up some new changes - we never use cldr instances as cache keys now, we just use locale strings, so all the caches are just plain objects now.

I also fixed the bug where `_resolveZoneIdText` was getting called instead of  `_cachedResolveZoneIdText`.

I also left a few comments above about memory consumption - everything seemed okay in my analysis (but can't say for 100% certainty that I was doing things correctly haha).

packages/locale/src/format/cldr/CldrCache.js
pithu: LGTM, so the the Cldr instance from external code seems to be harmless.

packages/locale/src/format/cldr/CldrCache.js
pithu: Since the normal use case should be only handfull of locales, it shouldn't be a problem. And even with hundreds of locales its just some MB, shouldn't harm a modern browser.

LGTM 👍

Issue #452, opened 2020-10-02
ES5 compatibility issues


scottlowe:
Hi there. Assuming that js-joda still intends to remain ES5 compatible, there is a call to `Array.includes` here:

https://github.com/js-joda/js-joda/blob/a1a635a7e4e506dfa7a9eedcc932571bdba9c1b2/packages/locale/src/format/cldr/CldrDateTimeTextProvider.js#L226

... which breaks compatibility with ES5 browsers (in my case IE11).

There is also a call to `Object.values` somewhere that is causing the same ES5 issue elsewhere in the code; I think that it might be this line:

https://github.com/js-joda/js-joda/blob/e663b249e2987784b60ebb22054d685ad13ed443/packages/locale/src/format/LocaleStore.js#L54

pithu:
Hi @scottlowe,

thanks for using js-joda and reporting this issue !

since we promised to be compatible with ie9, i will fix it. Unfortunately we had to disable ie test on travis for some reasons, therefore issues like that can sneak in.

scottlowe:
Woo! 😄

PR #453, opened 2020-10-08
Fix es5 issues

pithu:
Fix for es5 compatibility.

Fix for #452


pithu:
Contains wrong commits, sorry

PR #455, opened 2020-10-15
Iso formetters

akonior:


pithu:
Hi @akonior, 

thanks a lot for using js-joda and for contributing. 

PR LGTM

PR #456, opened 2020-10-18
Implements localTime.atOffset

InExtremaRes:
This PR primarily implements `LocalTime.prototype.atOffset()` with its corresponding types. It also adds the missing type declarations for `Instant.prototype.atOffset()` and `LocalDate.prototype.atTime(OffsetTime)`.

The first commit is just a little homogenization of the `js-joda.d.ts` file. Basically it adds missing semicolons, change the order of some methods and remove some methods that were already declared in a base class; I add that here since is a very small change, without changing any behavior.

I add the generated files in `dist` folder after run `npm run build-dist`. If that commit doesn't belong here I can remove it.

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -1158,14 +1158,14 @@ export class Instant extends Temporal implements TemporalAdjuster {
     isAfter(otherInstant: Instant): boolean;
     isBefore(otherInstant: Instant): boolean;
     isSupported(fieldOrUnit: TemporalField | TemporalUnit): boolean;
-    minus(amount: TemporalAmount): Instant;
     minus(amountToSubtract: number, unit: TemporalUnit): Instant;
+    minus(amount: TemporalAmount): Instant;
     minusMillis(millisToSubtract: number): Instant;
     minusNanos(nanosToSubtract: number): Instant;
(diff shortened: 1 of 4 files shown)


phueper:
i think it's best to remove the commit that updates the dist files, we'll do that as part of the publishing steps

other than that... LGTM

i'm a bit confused why the `atOffset` has been commented, but it seems to be like that basically from the initial implementation, maybe OffsetTime was missing at that point and it never made it back in :shrug: 

(oh, and just to be sure, is that the PR you were mentioning in #389 as another candidate for the major version bump?)

InExtremaRes:
@phueper 

> i think it's best to remove the commit that updates the dist files, we'll do that as part of the publishing steps

Ok, I'll remove the commit and force push. 

> i'm a bit confused why the `atOffset` has been commented, but it seems to be like that basically from the initial implementation, maybe OffsetTime was missing at that point and it never made it back in 

It seems so. Offset(Date)Time was implemented this year in #416 but maybe these methods were overlooked.

> (oh, and just to be sure, is that the PR you were mentioning in #389 as another candidate for the major version bump?)

Nope. I notice these while working in those. I decided to open this PR first since these changes are simpler. I think I'll have a PR in some hours.

InExtremaRes:
@phueper Ready!

PR #457, opened 2020-10-19
Methods .equals are now type predicates in TypeScript

InExtremaRes:
This PR makes most of the `.equals` methods [type predicates](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) in the TS declarations.

A type predicate is a function that returns `boolean` but the return type is marked with the syntax `parameter is Type` and allows the type system to narrow the type of `parameter` to `Type` in the _then_ branch. In our case it allows patterns like this:

```ts
function foo(temporal: Temporal) {
    const localDate = LocalDate.of(2020, 12, 31);

    // temporal is any Temporal object, but...

    if (localDate.equals(temporal)) {
        // ...if this branch is taken we know that temporal must be
        // of type LocalDate (.equals also checks the constructor) and now
        // the type system knows it as well.

        // I can call a method of LocalDate without complaining
        return temporal.atStartOfDay();
    } else {
        // Here should go some different logic. In this branch temporal is
        // still of type Temporal.
    }
}
```

I convert to type predicates only the `.equals` of the most common classes.

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -1093,7 +1093,7 @@ export class Duration extends TemporalAmount {
     addTo<T extends Temporal>(temporal: T): T;
     compareTo(otherDuration: Duration): number;
     dividedBy(divisor: number): Duration;
-    equals(other: any): boolean;
+    equals(other: any): other is Duration;
     get(unit: TemporalUnit): number;
     isNegative(): boolean;
     isZero(): boolean;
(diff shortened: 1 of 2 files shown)


phueper:
@InExtremaRes thanks again.. i merged both #389  and #456 ... can you update this PR and then we can take a look at it?

InExtremaRes:
@phueper Ready to be reviewed.

PR #460, opened 2020-10-20
Makes Temporal a better base class, renaming methods and with default behavior

InExtremaRes:
This PR makes `Temporal` a base class with a more useful default implementation, removing `DefaultInterfaceTemporal` and renaming methods that are implementation details.

As previously discussed, JS doesn't have proper interfaces and this library uses abstract classes instead which limites to one base class. Before this PR `Temporal` already had some common behavior but most were on `DefaultInterfaceTemporal`, in contrast to `TemporalAccesor`. Also most `Temporal`-derived classes repeated the same code to plus and minus a `TemporalAmount`.

The concrete list of changes:
- The class `DefaultInterfaceTemporal` is removed and its implementation moved directly into `Temporal`, making `Temporal` a most useful base class.
- Methods `plus`, `minus` and `with` of `Temporal` are semantically not abstract and shouldn't be overriden very often, probably never. This cannot be enforced in a simple way in JS/TS but the TS types now declare them as concrete methods. The default implementation doesn't change and still dispatch to different function depending on arguments (as `with` that can be called with an adjuster or value-unit pair).
- The functions to which the previously mentioned methods dispatch to are now prefixed with an underscore to clearly express that are implementation details and are not meant to be called explicitly. These methods are now called `_minusUnit`, `_minusAmount`, `_plusUnit`, `_plusAmount`, `_withAdjuster` and `_withField`; derived classes can override them to customize the behavior.
- Of the previously mentioned methods only `_plusUnit` and `_withField` are abstract and MUST be implemented, the rest have a sane default behavior taken even from `DefaultInterfaceTemporal` or the classes themselves. Methods `isSupported`, `until`, and `getLong` (inherited from `TemporalAccesor`) still are abstract and MUST be implemented.

The rationales for making this changes:
- Moving towards cleaner code, with less duplication and clearer concepts. This is just a little step but forward nonetheless.
- Helping users to distinguish methods that are not part of the public API (underscor-prefixed methods are common for this).
- Helping third-party implementation of `Temporal` (I guess not that many) to know which methods must be implemented and providing default behavior in most cases.

This is clearly a breaking change. Users previously using methods like `plusAmount` should now always use `plus` and so on with the rest. Users extending `Temporal` will need to change the names of the methods the're overriding and probably just implementing  `_plusUnit` and `_withField`.


InExtremaRes:
While this PR can still be reviewed I think I can remove a little more duplication, do a cleanup and make a couple of additional tests. I'll be pushing these days.

packages/core/src/temporal/Temporal.js
```diff
@@ -148,9 +152,13 @@ export class Temporal extends TemporalAccessor {
...
      */
-    // eslint-disable-next-line no-unused-vars
-    minusAmountUnit(amountToSubtract, unit) {
-        abstractMethodFail('minusAmountUnit');
+    _minusUnit(amountToSubtract, unit) {
+        requireNonNull(amountToSubtract, 'amountToSubtract');
+        requireNonNull(unit, 'unit');
+        requireInstance(unit, TemporalUnit, 'unit');
+        return (amountToSubtract === MIN_SAFE_INTEGER
+            ? this._plusUnit(MAX_SAFE_INTEGER, unit)._plusUnit(1, unit)
```
InExtremaRes, line 160: This code was taken from `DefaultInterfaceTemporal` which in turn was taken from the ThreeTen, but I think is incorrect in JavaScript.

In Java this equality holds:

```
-(Long.MIN_VALUE) == Long.MAX_VALUE + 1
```

But in JS the next is true instead:

```
-(Number.MIN_SAFE_INTEGER) == Number.MAX_SAFE_INTEGER
```

So adding 1 more unit is wrong here.
Besides, in Java a `long` (or any integer) just go back to its minimum value in an overflow situation, and that is the expected behavior of the original code I think, but in JS the `number` will just increase beyond `MAX_SAFE_INTEGER` (until it reaches `Number.MAX_VALUE` or `Infinity`):

```
Number.MAX_SAFE_INTEGER + 1 > Number.MAX_SAFE_INTEGER
```

The code should at least be modified to remove the `._plusUnit(1, unit)` part (and in any other places where that's relevant). If the code should try to emulate the integer-modular semantic of Java is something I'm not sure.

There is a new commented test that right now fails for this code.

pithu:
@InExtremaRes do you have anything else for a major release. @phueper and me have been very busy in recent times, sorry for that, but finally i have some time and plan to make a major release soon.

InExtremaRes:
@pithu I've been away from this project lately but I might have something in progress. Give me a couple of days and I'll reply to you here to let you know. I remember being working in more cleanup but those changes were dependent on this ones, so I didn't continue at that time.

PR #463, opened 2020-11-19
initial work to avoid need for 'init' fn

henryw374:
I would like to avoid the need for the init function to set properties on classes. This makes js-joda more amenable to static analysis, dead code elimination & etc. In my case I am using the google closure compiler. 

To avoid doing what the init function does, I am mostly using static getters. 

This approach seems to be working generally - although not yet finished. I haven't yet done Offset* classes.

Are you interested in this? If so please let me know. 

Thanks


pithu:
Hi @henryw374 

Thanks a lot for providing this very interesting approach to get rid of this nasty static _init initializer, but to be honest i wonder if this approach will work. Here some questions that came in my mind.

In the static getter, you delete the previous value. Wouldn't that mess up comparison of values, because new instances created over and over? eg `Month.JANUARY === Month.JANUARY` would be false.

The most difficult thing is to initialize the static properties in the right order and bring the import order in the right order. That's because of all this dependency cycles in threeten and javascript is transpiling directly when importing. 
Therefore, if things are not in the right order you can run in this `ReferenceError`: `ClassXY is not defined` errors. 
And with static getters, could that even lead to issues like this during runtime, depending in what order someone use static properties in his code base ? If so, we might still need a static initializer, that runs over all static props and initially access them once.

I doubt that all this would help with dead code analyzes and tree shaking, because i would guess the problem are the dependency cycles of the import statements. Or do i oversee something?

When i do an `npm install` on this branch, i get an compiler warning ? Fixing this error and then run `npm test` fails immediately, because of one of this nasty `ReferenceError`, in my case its `ZoneIdFactory is not defined`. Did you commit everything, is that working for you ?

Best Regards

henryw374:
Hi,

Thanks for the feedback. Sorry the branch is not working as-is - I have just been using it with google Closure and to do that, I've had to put all the module contents into a single file first. Since you are interested.. I'll have a go at making the normal tests work. 

First, regarding the `delete`, this was a technique I think I got from MDN somewhere. On the first call, the getter replaces itself with a property. So the next time it is referenced, the property is returned and the function has gone. See this dummy example. `Foo.bar === Foo.bar ` is true.

```
class Foo {

  static get bar() { 
      const result = {x:1 }; 
      delete Foo.bar; 
      Foo.bar = result; 
      return result}
}
```

Regarding the order, as I say I have been putting all the contents, minus the imports into a single file... but hopefully the approach works with normal imports - might need a bit of tweaking still... I'll have a go and see. 

In theory Google closure analyzes the whole program and so if the only js-joda fn I call is `Duration.parse`  then only that fn (and whatever other fns that calls) will get put into the program output. But.. the results aren't as good as I hoped.. in many cases with js-joda fns I call, it ends up dragging in a lot of js-joda that is not being called... and it doesn't explain why. I guess it just has to play safe. Still, if I import js-joda and don't call anything of it, then it is elided from the output... so that could be helpful for a library that depends on js-joda. 

And because the results aren't totally what I hoped for, having put in some effort back in the summer, I have shelved this for a while. But I thought it's worth putting a bit more time in to see if I can get this branch working so I'll try and do that in the near future and update again here if it looks ok.

henryw374:
I've had a go at getting this working with the normal `npm test` and yes the problem is getting ReferenceError. The way I am using the code in my branch is first putting all file contents into a single file, stripping off the import statements  - that way I don't get reference errors. So... I guess js-joda would have to go down the route of many classes per file if init is to be avoided in the normal build. I don't know if a single file would be required or there are just a few that would need combining.

PR #529, opened 2021-08-20
Please support Korean prebuilt

leejaycoke:
Thanks


pithu:
Sorry for the late reply, i will have a look later this week

PR #542, opened 2021-09-02
Remove edge case handling for minusUnit

pithu:
Handle this case as in all other temporal implementations.

Not sure why threeten is handling the edge case in such a strange way, but since js-joda is currently using the default implementation just for `Year` and `YearMonth`. let's ignore the edge case, makes no sense.


phueper:
hmm... could this be some strange case of signed int handling that is needed in java? anyway ... lgtm :)

phueper (approved):

PR #544, opened 2021-09-02
Add some more predefined Locales

pithu:


zawodskoj:
Hi, is it possible to also add russian locales?

PR #547, opened 2021-09-21
Fixes #546 by reverting .equals() being type predicates

### LINKED ISSUE #546: Type predicates break TypeScript semantics (PR #457)
The changes introduced in #457 break TypeScript semantics because `.equals()` isn't really a type predicate. Just because two temporals are unequal to each other doesn't mean that the other one has a different type.

Simple example:
```typescript
function issue(): void {
    const today = LocalDate.now();
    const someday = LocalDate.of(2021, 9, 20);
    if (today.equals(someday)) {
        console.log(someday.year());
    } else {
        // This now raises a compiler error because TypeScript thinks that
        // someday has type never.
        // error TS2339: Property 'year' does not exist on type 'never'.
        console.log(someday.year());
    }
}
```

InExtremaRes:
Fixes #546 

It (mostly) reverts #457, the tests were updated so they now fails when .equals are type predicates (a revert would have simply removed the tests).

### DIFF
--- a/packages/core/dist/js-joda.d.ts
+++ b/packages/core/dist/js-joda.d.ts
@@ -1103,7 +1103,7 @@ export class Duration extends TemporalAmount {
     addTo<T extends Temporal>(temporal: T): T;
     compareTo(otherDuration: Duration): number;
     dividedBy(divisor: number): Duration;
-    equals(other: any): other is Duration;
+    equals(other: any): boolean;
     get(unit: TemporalUnit): number;
     isNegative(): boolean;
     isZero(): boolean;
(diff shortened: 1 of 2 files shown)


pithu (approved):
LGTM, i am not sure if we really need the `not a predicate` tests, it feels a bit strange to test for things not to be. But they do not harm.

InExtremaRes:
@pithu The tests are a minimal reproduction of the report in #546. I agree they feel a little strange but I thought they could serve as regression tests in case anyone else (including a future me :smile:) try to implement a type predicate again. If you prefer I can remove them.

pithu:
> @pithu The tests are a minimal reproduction of the report in #546. I agree they feel a little strange but I thought they could serve as regression tests in case anyone else (including a future me 😄) try to implement a type predicate again. If you prefer I can remove them.

no, fine for me, thanks for contributing

Issue #157, opened 2017-05-07
LocalDate to LocalDate comparison via opertors


paplorinc:
Hey,

We're trying out `js-joda` as ` moment` alternative for `TypeScript`. So far we like it, and have a question.

Knowing that these return the expeced results:
```typescript
LocalDate.of(2017,1,10) < LocalDate.of(2017,1,10) // false
LocalDate.of(2017,1,10) <= LocalDate.of(2017,1,10) // true
LocalDate.of(2017,1,10) > LocalDate.of(2017,1,10) // false
LocalDate.of(2017,1,10) >= LocalDate.of(2017,1,10) // true
LocalDate.of(2016,1,10) < LocalDate.of(2017,1,10) // true
```
etc.
(Note, `==` or `===` doesn't seem to be working however...)

Is it valid to compare `LocalDate`s and the rest via operators, as it seems to be working correctly, but couldn't find it in the docs (or even where the `valueOf` was overridden to allow this behavior).

Thanks,
Lőrinc

phueper:
Hi @paplorinc,

thanks for reporting and for using js-joda. 

To be honest, i haven't thought about using operators with joda objects, it might be a good idea to use / implement `valueOf` for the joda date objects, but that is currently not the case. 

The fact that the comparison using `<`, `>`, `<=`, `>=` works is very probably coincidence, maybe has something to do with the joda objects implementing `toString` in a way that a comparison might work? But i'm really not sure...

In summary... i think using operators to compare joda Objects is an interesting idea, would be a topic for the roadmap but is currently not in our focus. To compare js-joda objects you should use the `compareTo` and `equals` methods, e.g. https://js-joda.github.io/js-joda/esdoc/class/src/LocalDate.js~LocalDate.html#instance-method-compareTo ... 

Regards, Pattrick

augusthur:
Equals (==) operator is necessary if you want to use Maps or Sets.

For example:
```
let da = LocalDate.parse("2012-12-24");
let db = LocalDate.parse("2012-12-24");
let dateSet = new Set();
dateSet.add(da);
dateSet.add(db);
dateSet.size // will be 2, storing both identical dates.
```

A workaround is to store dates as strings but makes code less readable.

pithu:
Hi @paplorinc 

as far as i understand the es6 spec that's a problem with the `Set` specification http://www.ecma-international.org/ecma-262/6.0/#sec-set.prototype.add. 

[...]

PR #548, opened 2021-09-21
Disallowing implicit conversion of Temporal/TemporalAmount to numeric values

InExtremaRes:
This PR is an experiment that implements a function `Temporal.prototype.valueOf` that always throws as discussed in #157.
If the idea is accepted I still think the same should apply to `TemporalAmount` and maybe `TemporalAccessor`.

As you can see in the changes any implicit coercion throws, including something like `temporal + "string"`. The reason is that operator `+` calls `valueOf()` if it exists and only calls `toString()` if it doesn't, even if the RHS argument is a string. In those cases is necessary to even explicitly convert to string (using `.toString()` or `String()`) or use a string context (like strings using backticks).

What are your thoughts on this?

**EDIT:**
Finally the decision was not to use `valueOf` but `Symbol.toPrimitive` so implicit coercion to string is still allowed. This has some implications:

- It doesn't work in `< ES6`, so the coercion won't throw in ES5 environments. Since those should be a minority it was decided to just go with it.
- Adding two `Temporal`s together (or `TemporalAmount`) using operator `+` won't throw, instead both sides of the operation will be converted to `string` and concatenated. This includes, for instance, `LocalDate.now() + 99` which will yield something like `"2021-10-3199"`. This is the usual behavior in EcmaScript; hardening it will prevent handy concatenation between `string` and `Temporal` objects, which might be too much of a breaking change.
- It will throw, however, in cases like using operators `<`, `>=` or the unary `+`.


packages/core/src/DayOfWeek.js
```diff
@@ -93,7 +93,7 @@ export class DayOfWeek extends TemporalAccessor {
      */
     static of(dayOfWeek) {
         if (dayOfWeek < 1 || dayOfWeek > 7) {
-            throw new DateTimeException('Invalid value for DayOfWeek: ' + dayOfWeek);
+            throw new DateTimeException('Invalid value for DayOfWeek: ' + dayOfWeek.toString());
```
pithu, line 96: Why add a .toString() here? Is the method implemented for all types ? And isn't toString used anyway ?

packages/core/src/Month.js
```diff
@@ -572,7 +572,7 @@ export class Month extends TemporalAccessor {
             return Month.of(temporal.get(ChronoField.MONTH_OF_YEAR));
         } catch (ex) {
             throw new DateTimeException('Unable to obtain Month from TemporalAccessor: ' +
-                    temporal + ' of type ' + (temporal && temporal.constructor != null ? temporal.constructor.name : ''), ex);
+                    String(temporal) + ' of type ' + (temporal && temporal.constructor != null ? temporal.constructor.name : ''), ex);
```
pithu, line 575: And why using here String ? Nowadays i would prefer template literals. WDYT ?

packages/core/src/temporal/Temporal.js
```diff
@@ -350,4 +350,14 @@ export class Temporal extends TemporalAccessor {
...
+
+    valueOf() {
+        throw new TypeError(
+            'Conversion to a primitive type is not allowed. ' +
+            'If you are trying to compare using operators `<`, `=`, etc., ' +
+            'please use the methods `equals()`, `compareTo()` or the one ' +
+            'that better suite your use case. If you are trying to convert to ' +
+            'string use the `toString()` or `format()` methods instead.'
+        );
+    }
```
pithu, line 362: 👍

packages/core/test/reference/InstantTest.js
```diff
@@ -1710,20 +1710,20 @@ describe('org.threeten.bp.TestInstant', () => {
...
                     const b = instants[j];
                     if (i < j) {
-                        assertEquals(a.compareTo(b) < 0, true, a + ' <=> ' + b);
-                        assertEquals(a.isBefore(b), true, a + ' <=> ' + b);
-                        assertEquals(a.isAfter(b), false, a + ' <=> ' + b);
-                        assertEquals(a.equals(b), false, a + ' <=> ' + b);
+                        assertEquals(a.compareTo(b) < 0, true, `${a} <=> ${b}`);
+                        assertEquals(a.isBefore(b), true, `${a} <=> ${b}`);
+                        assertEquals(a.isAfter(b), false, `${a} <=> ${b}`);
+                        assertEquals(a.equals(b), false, `${a} <=> ${b}`);
```
pithu, line 1716: 👍

packages/core/test/temporal/TemporalTest.js
```diff
@@ -113,4 +113,10 @@ describe('js-joda Temporal', () => {
             expect(new BasicYearMock(2).equals(result)).to.equal(true);
         });
     });
+
+    describe('valueOf', () => {
+        it('should throw an exception', () => {
+            expect(()=> new BasicYearMock(1).valueOf()).to.throw(TypeError);
+        });
+    });
```
pithu, line 121: Does it make sense to add a coercing test here as well ?

pithu (commented):
Hi great that you finally work on this! In general i like the PR, I just have some question.

packages/core/src/Month.js
InExtremaRes: In general I also prefer template literals. In the changes I made I kept the code as it was (except in tests) just adding the necessary parts to keeping it working. Here I used `String` instead of `.toString()` because this method is called with `null` in a test and `String()` works with `null` and `undefined` while `.toString()` doesn't.

Do you think is better to convert this to templates literals instead?

packages/core/test/temporal/TemporalTest.js
InExtremaRes: My reasoning for this is that `BasicYearMock` is a minimal implementation of `Temporal`, so this tests that the behavior is really inherited for every `Temporal` and not only to the "standard" ones (like `LocalDate`, `Instant`, etc).

packages/core/src/temporal/Temporal.js
InExtremaRes: Do you think the wording is right? English is not my first language and I still make silly mistakes.

packages/core/test/reference/InstantTest.js
InExtremaRes: :smiley:

packages/core/src/DayOfWeek.js
InExtremaRes: Indeed, this is not needed. I put it because I was first thinking in implement this method for every major type; at the end I just make a first experiment only with `Temporal` to see how it goes and discuss about it.

What do you think? What types should have the same behavior? Only `Temporal` and `TemporalAmount`? All of them?

pithu: Not sure if i get your answer right. The `.toString()` is not required, because `toString()` method that is automatically called when the object is to be represented as a text value. And it is implemented for all threeten types.
And yes, IMO it would be nice to convert to string literals, whenever we touch a piece of code. so ``Invalid value for DayOfWeek: ${dayOfWeek}`` should do the trick

packages/core/src/Month.js
pithu: Not sure if i get it, isnt it the same as ``${temporal} of type ${temporal && temporal.constructor != null ? temporal.constructor.name : ''}``

packages/core/src/temporal/Temporal.js
pithu: My english is terrible as well, i would guess its fine, just `... or one that is more suitable for your use case. ...`

pithu (commented):
Looks like tests in extra package are failing with your changes

packages/core/src/DayOfWeek.js
pithu: And yes, i would say we should also throw in `TemporalAmount.valueOf`

InExtremaRes: Sorry I wasn't clear. In the expression `"text" + o` where `o` is an object, the method `o.valueOf()` is called even if the LHS is a string; `.valueOf()` defaults to `.toString()` if isn't implemented (there is also `Symbol.toPrimitive` but let us left that aside for now). Even if an object has an implementation for `.toString()` it is not called if there is an implementation for `.valueOf()`. OTOH a template literal string does call `.toString()` first and not `.valueOf()`.

When I started this I was thinking to implement a throwing `valueOf` for almost every type, but then just did for `Temporal` to see how it goes. This particular line was the "fix" for a `.valueOf()` that throws in `DayOfWeek` but at the end I submit a draft PR to discuss about this before implementing more. If we decided that `DayOfWeek` won't have that method then I will revert this line, if it should throw as well then this `.toString()` is necessary (or changing it to a template literal).

packages/core/src/Month.js
InExtremaRes: Yes, changing to a template literal is clearer. I just "fix" the tests by putting a bunch of `.toString()` where was needed. I couldn't use `.toString()` in this particular case because there is a test that calls this function with `null` and can't call `.toString()` with `null`, causing the test to fail. Using `String()` instead was the easy fix.

InExtremaRes:
Thank you for taking a look. If you think this change is a good approach I will address the comments, clean the code and take a look at the failing tests. I'll try to change the concatenated strings to template literals. I'll also see how it goes doing the same change to `TemporalAmount`.

I'm super busy at work these days but will try to make this changes during this week.

packages/core/src/DayOfWeek.js
pithu: Ok, that changes things a bit. That would mean, this PR breaks a lot, correct ?

A `console.log('Today is ' + LocalDate.now());` would then throw an exception, because `valueOf()` is called internally. In opposite ``console.log(`Today is ${LocalDate.now()}`);`` would not throw, because it fallbacks to `toString()`

Hmm, i have to think about it, not sure if this is what we want ? 

I just checked temporal/proposal polyfill, they have the same issue. `console.log('foo' + Temporal.Now.plainDateISO());` throws a TypeError.

pithu: @phueper any opinion on that ? 

In general it's about: throwing a Typeerror on any comparison via operators, so that a user can't perform an invalid comparison, eg `LocalDate.of(2017,1,10)  < LocalDate.of(2017,1,10)` would throw instead of returning false. 

In consequence that is a major change, that disables String concatenation as well.

phueper: my thoughts:
- in general i do like the idea of throwing for example when comparing
- throwing when doing `console.log('Date:' + LocalDate.now())` however seems a little risky to me .. i would expect a lot of existing code to break with that ( i can remember me doing this several times :)) but i assume there is no way for us to know wether valueOf is called for this kind of string conversion or for another operation?

pithu: Is there an autofix option for string literals in eslint ?

pithu:
> As far as I know there is no way to differentiate between the two, at least no in ES5. In ES6 was introduced Symbol.toPrimitive that allows to implement a function that takes a hint and might help here. But again, not in ES5.

Our code is bundled to ES5 code, so even template literals are converted to string concats with the plus operator (see packages/core/dist/js-joda.js). So not even our bundled  code would work anymore, because any internal string concat would throw a type error. Correct ?

If so, why not switch to Symbol.toPrimitive, would solve the issue for es6, and that is the majority of all environment. eg the following impl would throw for `date1 > date2` but not for `date1 + date2`. In es5 environments everything would be as before.

```
class Temporal {
    // ...
    [Symbol.toPrimitive](hint){
        if (hint === 'number') {
            throw new TypeError('use isAfter(), isBefore or isEquals() to compare Temporals');
        }
        return this.toString();
    }

```

pithu:
And i would advocate for extending our eslint rules by `'prefer-template': 'error',`. The rule enforce the use of string template and can auto fix our existing code. (i just checked that)
WDYT ?

phueper:
> 
> If so, why not switch to Symbol.toPrimitive, would solve the issue for es6, and that is the majority of all environments. eg the following impl would throw for `date1 > date2` but not for `date1 + date2`. In es5 environments everything would be unchanged.
> 
> ```
> class Temporal {
>     // ...
>     [Symbol.toPrimitive](hint){
>         if (hint === 'number') {
>             throw new TypeError('use isAfter(), isBefore or isEqual() to compare Temporals');
>         }
>         return this.toString();
>     }
> ```
> 
> WDYT ?

that sounds much better to me

InExtremaRes:
> Our code is bundled to ES5 code, so even template literals are converted to string concats with the plus operator (see packages/core/dist/js-joda.js). So not even our bundled code would work anymore, because any internal string concat would throw a type error. Correct ?

That's my understanding, yes.

> 
> If so, why not switch to Symbol.toPrimitive, would solve the issue for es6, and that is the majority of all environments. eg the following impl would throw for `date1 > date2` but not for `date1 + date2`. In es5 environments everything would be unchanged.
> 
> ```
> class Temporal {
>     // ...
>     [Symbol.toPrimitive](hint){
>         if (hint === 'number') {
>             throw new TypeError('use isAfter(), isBefore or isEqual() to compare Temporals');
>         }
>         return this.toString();
>     }
> ```
> 
> WDYT ?

Yes. Actually I was thinking something similar these days and I think I like the idea, but there is still a gotcha: in ES5 there is no `Symbol` object so `Symbol.toPrimitive` is a TypeError. What I was thinking was conditionally modify `Temporal.prototype` depending if `Symbol.toPrimitive` is a thing; that shouldn't be any more complicated.

Is there any test in place for ES5 environments?

InExtremaRes:
BTW Is there a reason to keep supporting ES5? Just curious.

pithu:
> BTW Is there a reason to keep supporting ES5? Just curious.

Good question, ie11 usage is about 1%, not much but still some user around, so IMO we need good reason to break with es5, if there is a work around we should try.

pithu:
> Is there any test in place for ES5 environments?

Another good question, we used to have saucelabs browser tests for several environments, among other for `windows 10, internet explorer`. But we skipped most of the saucelabs tests, because they have been flaky.

Anyway, the tests that run against code inside the same package, do not test the webpack bundled code. Just test cases that do import classes from other packages do so, eg `packages/extra/test/reference/IntervalTest.js` is testing bundled code from `@js-joda/core`. 
So a trick to tests this, could be to activate our saucelab profile `sl_ie` and write tests in `extra` package that test code in `core`. Could work, but quit a hack. Or another more proper way would be to write such tests in the `example` package.

For manual tests, you can use vm's from microsoft eg from here https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/. I am curious now, and will do some manual tests on ie11 with `Symbol.toPrimitive`

InExtremaRes:
I been very short on time these days. I have this idea: I'll make a new commit (mostly) reverting this and making the changes I have in mind to be reviewed again. At the end I can just open a new, clean PR or something.

pithu:
@InExtremaRes I am just about to switch the whole lib to prefer template eslint rule with #551, so please sync as soon as merged. I hope that will not cause conflicts for you

pithu:
FYI: 
I did some more cleanups, i am done now and will publish a new minor release soon. 
- I did some documentation polish. A.o. the latest js-joda build is injected into the browser dev console now, good for testing, eq on ie11
- I cleaned up the webpack bundle.
- I fixed some issues with ie11 and manually tested it, should work again
- Automated test on saucelabs running now for ie11.

InExtremaRes:
The test are failing but the error seems to suggest is a problem with SauceLabs. Could you take a look?

InExtremaRes:
I'm very sorry about the delay, I was super busy the last few weeks.

I made the same change now for `TemporalAmount` so this PR should be ready for the final review. I still want to put some words in the documentation but I'm thinking I'll do that in a separate PR so this one is not posponed any longer.

I haven't tested this in ES5. If you have an environment ready I'd be very grateful if you can test it.

pithu (approved):
LGTM, i manually tested with ie11 works in ie11, so in ie11  implicit conversion still is possible.

Let me know when you are done, then i would merge the PR

InExtremaRes:
> Let me know when you are done, then i would merge the PR

I think it is. I still want to add a few words to the documentation but I'll do that in another PR. If you see that everything is ok, then can be merged :)

PR #564, opened 2021-10-13
Fix travis for PR's from forks

pithu:
Motivation for this PR and recent cleanups of the ci pipeline:

- Saucelabs tests require credentials, for security reasons they can't be used in forked pull requests, therefore must be skipped in that case. A recent cleanup broke this.
- Travis has changed its pricing plans. OpenSource projects are still free, but you only get a few credits on request and as soon as the credits run out, you have to manually contact Travis support and ask for more credits. This is really annoying.
- GitHub actions free pricing plan includes 2000 minutes per month. Not really better, because if you run out of credits you can't run jobs anymore until the end of the month. Times getting harder for OpenSource projects.
- Sauce connect was not really stable on travis
- Phantomjs is deprecated and should be replaced by headless chrome and firefox test

So what is the new ci setup:

- I split the the ci jobs in a travis part and a GitHub actions part.
   - Travis is running on push to master and on every PR, also for pull requests from forked repositories
   - Travis is testing with node.js, doing the the code coverage and karma tests with headless firefox and chrome
   - I removed the step matrix from travis. That increased the time a bit until all jobs are done, 
     but the total runtime of all steps together is about the half. That means we save credits.
   - GitHub actions is running on push to master and on every PR from a non forked repositories.
   - GitHub actions is just performing the sauce labs test
- So by splitting the ci jobs we increased our total ci job budget, let's keep fingers crossed that its enough
- In GitHub its more easy to control when the sauce connect is running, let's hope its more stable.

@InExtremaRes After this PR is merged and you rebase / merge your branch with it, your ci issues should be solved.

### DIFF
--- a/.github/workflows/saucelabs-test.yaml
+++ b/.github/workflows/saucelabs-test.yaml
@@ -14,12 +14,6 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.push || github.event.pull_request.head.repo.full_name == github.repository
     steps:
-#      - name: dump github context
-#        env:
-#          GITUB_CONTEXT: ${{ toJson(github) }}
-#        run: echo "$GITUB_CONTEXT"
-#      - name: dump default env
-#        run: export
       - name: Checkout
         uses: actions/checkout@v2
       - name: Setup node 14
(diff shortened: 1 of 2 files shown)


phueper:
lgtm ... too bad, that the CI Setup gets more complicated (no single CI but now two) ... but i like the switch to github actions... maybe in the future we might even be able to do auto releases (on tags?) from github actions?

phueper:
also... can we track how many credtits/minutes we use per month ? maybe we can just switch to github actions completely, if the free minutes are enough ...

pithu:
> lgtm ... too bad, that the CI Setup gets more complicated (no single CI but now two) ... but i like the switch to github actions... maybe in the future we might even be able to do auto releases (on tags?) from github actions?

Yes switching to github actions should be fairly easy. To auto-release / publish we probably have to remove all the generated distribution files (at folder `dist`) from git and just generate it before publishing. That would also imply that we have to generate docs during the publish step and commit it.

PR #574, opened 2021-12-26
Reverting babel targets

pithu:
This PR reverts the babel targets to the previous config (latest two versions / UMD). The current config mess up the transpiled code, make it very slow and memory consuming. 

The config of this PR is running with win8/ ie11, but there is a strange issue JSJodaLocale global var. It might be a bug, that needs to be fixed. The core, extra and timezone packages are working fine with ie11. On win10/ MSEdge everything seems to work. 

So since ie11 is rarely used and the issue seems to be fixable, i guess we should revert to the previous config


phueper:
lgtm

phueper (approved):

PR #577, opened 2021-12-28
npm update dependencies

pithu:


pithu:
Will be fixed in #579

PR #578, opened 2021-12-30
WeekFields export and typings

StrayAlien:
exporting WeekFields and ComputedDayOfField from locale package.  Plus, typings for same.  Permits one to do: 

```
import {WeekFields} from "@js-joda/locale";
```

rather than import directly from locale/src folder (as per docs) and then have to juggle your ts/babel/jest/etc config to accomodate that one file ... 

I've added a typing for `EnumMap`.  Unsure if you'd like this in the core typings, but, it is in the locale typings atm.  Let me know if you want that changed.

### DIFF
--- a/packages/locale/src/js-joda-locale.js
+++ b/packages/locale/src/js-joda-locale.js
@@ -5,12 +5,11 @@
 import { use } from '@js-joda/core';
 import plug from './plug';
 import Locale from './Locale';
-import { WeekFields, ComputedDayOfField } from './temporal/WeekFields';
+import { WeekFields } from './temporal/WeekFields';
 
 use(plug);
 
 export {
     Locale,
-    WeekFields,
-    ComputedDayOfField
+    WeekFields
 };
--- a/packages/locale/src/js-joda-locale.js
+++ b/packages/locale/src/js-joda-locale.js
@@ -5,9 +5,12 @@
 import { use } from '@js-joda/core';
 import plug from './plug';
 import Locale from './Locale';
+import { WeekFields, ComputedDayOfField } from './temporal/WeekFields';
 
 use(plug);
 
 export {
-    Locale
+    Locale,
+    WeekFields,
+    ComputedDayOfField
 };
--- a/packages/locale/typings/js-joda-locale.d.ts
+++ b/packages/locale/typings/js-joda-locale.d.ts
@@ -16,8 +16,8 @@ export class WeekFields {
     public static ISO: WeekFields;
     public static SUNDAY_START: WeekFields;
     public static ofLocale(locale: Locale): WeekFields;
-    public static of(firstDayOrLocale: core.DayOfWeek | Locale, minDays?: number);
-    public static ofFirstDayOfWeekMinDays(firstDayOfWeek: DayOfWeek, minimalDaysInFirstWeek: number);
+    public static of(firstDayOrLocale: core.DayOfWeek | Locale, minDays?: number): WeekFields;
+    public static ofFirstDayOfWeekMinDays(firstDayOfWeek: DayOfWeek, minimalDaysInFirstWeek: number): WeekFields;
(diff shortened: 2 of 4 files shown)


pithu:
Hello @StrayAlien 

thanks for this PR, i guess exporting WeekFields makes sense. I am unsure with ComputedDayOfField, because its not part of the threeten API. What is the use for ?

@phueper WDYT ?

StrayAlien:
Thanks @pithu - tbh, it was arbitrary as it was in the same code unit - but ... I see that WeekFields does not actually expose it, so, I have removed the ComputedDayOfField typing.  Keep it easy and also means the EnumMap type can go as well.

Also, in the latest commit above I've removed the typing for OffsetTime.toEpochSeconds().  Hope that is okay also.

packages/locale/src/js-joda-locale.js
```diff
@@ -5,9 +5,12 @@
...
 import Locale from './Locale';
+import { WeekFields, ComputedDayOfField } from './temporal/WeekFields';
 
 use(plug);
 
 export {
-    Locale
+    Locale,
+    WeekFields,
+    ComputedDayOfField
```
pithu, line 15: Its still exported ?

pithu (commented):
I would expect a very basic test for WeekFields at `packages/locale/test/typescript_definitions/js-joda-locale-tests.ts` similar to the other typing tests

packages/locale/src/js-joda-locale.js
StrayAlien: Thanks @pithu - tests in and export removed.  Hopefully the tests are of the sort your require.

packages/locale/typings/js-joda-locale.d.ts
```diff
@@ -10,6 +12,22 @@ export class Locale {
     public equals(other: any): boolean;
 }
 
+export class WeekFields {
+    public static ISO: WeekFields;
+    public static SUNDAY_START: WeekFields;
+    public static ofLocale(locale: Locale): WeekFields;
+    public static of(firstDayOrLocale: core.DayOfWeek | Locale, minDays?: number): WeekFields;
+    public static ofFirstDayOfWeekMinDays(firstDayOfWeek: DayOfWeek, minimalDaysInFirstWeek: number): WeekFields;
```
pithu, line 20: LGTM, great work, also the tests! 

I am just wondering if we should keep this consistent and use `core.DayOfWeek` instead of `DayOfWeek`. Would this make the import above obsolete ?

InExtremaRes, line 20: Some of these methods aren't public and shouldn't be typed, in particualr `ofLocale` and `ofFirstDayOfWeekMinDays`.

Also, I think the `of` method could be typed as an overload instead of using an union type. Right now the type accepts a call with a `Locale` _and_ a `number`, which isn't right.

InExtremaRes: I am also in favor of moving the `import * as core from '@js-joda/core'` statement to the top and replace the other import.

InExtremaRes (changes requested):
Hi and thank you for contributing. I left just a couple of comments but it looks good to me overall.

StrayAlien:
Thanks for comments. Getting there.  

I've updated to reflect those comments.  Thank you.  I used this as a reference - I hope that it is correct: https://www.threeten.org/threetenbp/apidocs/org/threeten/bp/temporal/WeekFields.html.

InExtremaRes:
Just one more little thing. Could you please put the `import` statement at the top of the file? Even if for TypeScript is not an issue in this case, It looks a little weird where it is now. Thank you.

StrayAlien:
done

pithu (approved):
LGTM

InExtremaRes (approved):
👍

Issue #545, opened 2021-09-11
Add Instant.ofMicro() instantiator


mattbishop:
I am working with Postgres database timestamps which have microsecond accuracy (1 microsecond = 1000 milliseconds). I'm using js-joda to work with it, but to create an instant from the timestamp, I have to convert like this:

```typescript
      // Long is a 64-bit long lib: https://www.npmjs.com/package/long
      const epoch = Long.fromString(posgresTimestampEpochNumber)
      const time = epoch.divide(ONE_THOUSAND)
      const micros = epoch.subtract(time.multiply(ONE_THOUSAND))
      const tsInstant = Instant
        .ofEpochMilli(time.toNumber())
        .plus(micros.toNumber(), ChronoUnit.MICROS)
```

It would be really nice to have something like:

```js
const tsInstant = Instant.ofEpochMicro(epoch.toNumber())
```

mattbishop:
Also, please consider `tsInstant.epochMicro()` to provide the epoch time in microseconds.

pithu:
We are open for PR's for that case

mattbishop:
Thanks for the nudge! I will work on a PR to add this method.

ayushkamadji:
PR is merged. Perhaps this issue should be closed?

mattbishop:
Should we wait for a release before closing?

pithu:
The PR have been released -> https://github.com/js-joda/js-joda/commit/d07153c38a3d0b2b29b8185e3daaac69c624d4a9

mattbishop:
My mistake, I forgot the Typescript definition for this method hasn't been released yet: https://github.com/js-joda/js-joda/commit/9617179fc33cb73d2ca7f713cd821e71dec6af7b

PR #584, opened 2022-01-17
Support for Instant micros -- ofEpochMicros, plus and minus micros.

mattbishop:
PR for https://github.com/js-joda/js-joda/issues/545

Some applications like Postgres work in microseconds and it is difficult to convert a microsecond time into an Instant directly. I also added methods to add and subtract micros so that class is balanced with the other time parts (seconds, millis, nanos).


pithu (approved):
LGTM, thanks for contributing

PR #590, opened 2022-01-27
Add Instant micro methods to Typescript definitions.

mattbishop:
These methods were added to js-joda recently, and the typescript definition should match.


pithu:
Hi @mattbishop thanks for contributing, could you please add a basic test at packages/core/test/typescript_definitions

mattbishop:
Thanks for the request, I added typings tests in the 'Instant' test suite. I also added in flow method declarations, though I could not find any tests for those.

pithu (approved):

PR #591, opened 2022-01-31
Fix formatting in simple parser example

kcsmnt0:
The `formatter` needs to be passed into the `parse` function in order to be used in the parsing operation.


pithu (approved):
Thx a lot for contributing

PR #592, opened 2022-02-09
Fix broken table in WeekFields.js~WeekFields.html

Newbie012:
Ignore this PR if this file is autogenerated. Currently the table [here](https://js-joda.github.io/js-joda/class/packages/locale/src/temporal/WeekFields.js~WeekFields.html) is a bit broken.


pithu:
Yes its auto generated, Can you try to fix in the source file WeekFields.js.
 
You can test/ generate locally with the command `npm run build-esdoc`

PR #599, opened 2022-03-18
prebuilt packages for new locales (pl, da, el)

mstawick:


pithu:
LGTM, thanks for contributing

mstawick:
Thanks for the great library

PR #606, opened 2022-05-05
Add a prebuilt en-gb locale

benlondon:
We'd really benefit from an `en-gb` prebuilt locale, please could you consider adding this?


pithu:
LGTM, thanks for contributing

PR #614, opened 2022-07-16
add DayOfMonth and DayOfYear

perceptron8:
I've reimplemented 2 classes from threeten-extra:
- https://github.com/ThreeTen/threeten-extra/blob/master/src/main/java/org/threeten/extra/DayOfMonth.java
- https://github.com/ThreeTen/threeten-extra/blob/master/src/main/java/org/threeten/extra/DayOfYear.java

Tests should be more or less complete. They are rewrite of:
- https://github.com/ThreeTen/threeten-extra/blob/master/src/test/java/org/threeten/extra/TestDayOfMonth.java
- https://github.com/ThreeTen/threeten-extra/blob/master/src/test/java/org/threeten/extra/TestDayOfYear.java

I've extended typings and typings test.

Whenever I could, I tried to keep this project's code style and conventions (function overloading pattern, `value()` instead of `getValue()` and so on). Code base is quite big, so I may have missed something less obvious. This is my first contribution to js-joda, so please, be forgiving. ;)

Happy reviewing!


pithu:
Great news @perceptron8, thanks a lot for contributing. Unfortunately it will take some time until I get to review it, please have some patience

packages/extra/src/MathUtil.js
```diff
@@ -0,0 +1,236 @@
...
+ */
+import { ArithmeticException } from '@js-joda/core';
+
+export const MAX_SAFE_INTEGER = 9007199254740991;
+export const MIN_SAFE_INTEGER = -9007199254740991;
+
+/**
+ * Math helper with static function for integer operations
+ */
+export class MathUtil {
```
phueper, line 14: hmmm... i dont like duplicating MathUtil.js ... i wonder wether we cant somehow import it from js-joda/core ... and yes, i now it isn't exported by default ... but duplicating seems wrong ... maybe we need a utils package or sth similar?

phueper (commented):
Quicker than i thought ... overall looks good to me ... one minor comment about formatting, and a bigger one about MathUtils.js ... maybe we can think of a way that doesn't need copying MathUtils ?

packages/extra/test/typescript_definitions/js-joda-extra-tests.ts
```diff
@@ -1,50 +1,117 @@
...
-  interval.isBeforeInstant(instant);
-  interval.isAfterInterval(interval);
-  interval.isBeforeInterval(interval);
-  interval.toDuration();
-  interval.equals({});
-  interval.hashCode();
-  interval.toString();
-
-  LocalDate.ofInstant(interval.end());
+    const instant = Instant.now();
```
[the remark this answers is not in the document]
perceptron8: I can restore original formatting, but then it will become inconsistent.

packages/extra/src/MathUtil.js
perceptron8: I don't like duplicating MathUtil.js too, but this also applies to `requireInstance`, `requireNonNull` and test.js. Currently there is nothing leaking into public API here, so how about refactoring this later?
Importing from js-joda/core doesn't seem to be possible. I would be happy to see some utils package(s) instead.

perceptron8:
I've spotted two `requireNonNull(field, 'temporal');` that need to be replaced with `requireNonNull(field, 'field');`. I'll fix this later today.

packages/extra/test/typescript_definitions/js-joda-extra-tests.ts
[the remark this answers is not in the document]
perceptron8: What i meant is that for some time (if reformatting commit goes after DayOf-related changes), js-joda-extra-tests.ts would contain inconsistent indentation.

I can also split reformatting into separate commit in a way that it "predates" DayOf-related changes, but this would require rebase and force-push, which I tried to avoid as this could make your work harder (you would have to start over review).

It's up to you, just let me know.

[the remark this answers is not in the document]
phueper: yeah.. i think you are right .. sounds like too much effort for this small of a change ... i just like to have separate commits just for reformatting and then the actual "content" commit ... but i'd say leave it as it is

packages/extra/src/MathUtil.js
phueper: good point as well ... let's leave it for later and use it as it is right now ...

pithu: Its possible to use MathUtil from the core package, see https://github.com/js-joda/js-joda/blob/328c57028f43aa078b0329e2bb104bcc98aa61f1/packages/locale/src/format/parser/LocalizedOffsetPrinterParser.js#L14 for an example. Could you try to use this one instead ?

packages/extra/test/reference/DayOfMonthTest.js
```diff
@@ -0,0 +1,602 @@
...
+
+    //-----------------------------------------------------------------------
+    // now()
+    //-----------------------------------------------------------------------
+    describe('now()', () => {
+        it('test_now', () => {
+            let test = DayOfMonth.now();
+            if (LocalDate.now().dayOfMonth() !== test.value()) {
+                test = DayOfMonth.now();
+            }
```
pithu, line 86: What's the purpose if the if clause? Doing the same assignment again, looks strange.

packages/extra/test/reference/DayOfYearTest.js
```diff
@@ -0,0 +1,524 @@
...
+
+    //-----------------------------------------------------------------------
+    // now()
+    //-----------------------------------------------------------------------
+    describe('now()', () => {
+        it('test_now', () => {
+            let test = DayOfYear.now();
+            if (LocalDate.now().dayOfYear() !== test.value()) {
+                test = DayOfYear.now();
+            }
```
pithu, line 87: same comment here

pithu (commented):
Overall LGTM, just some nit comments and we should not to duplicate MathUtil, should work for `extra` package as it does for the `locale` package

packages/extra/test/reference/DayOfMonthTest.js
perceptron8: See also:
- https://github.com/js-joda/js-joda/blob/master/packages/core/test/reference/LocalDateTest.js#L118-L132,
- https://github.com/js-joda/js-joda/blob/master/packages/core/test/reference/YearMonthTest.js#L37-L53.

Without mocking (freezing) `Clock.systemDefaultZone()` which `now()` uses, similar tests may be impossible to rewrite to be fully deterministic.

perceptron8: Fixed in 878336ab4c1a6fe2cf892bc20a1725bae8628904 (still non-fully-deterministic, but at least correct).

packages/extra/test/reference/DayOfYearTest.js
perceptron8: Fixed in 878336ab4c1a6fe2cf892bc20a1725bae8628904.

packages/extra/test/reference/DayOfMonthTest.js
```diff
@@ -80,11 +80,16 @@ describe('org.threeten.extra.TestDayOfMonth', () => {
...
+            let actual;
+            for (let i = 0; i < 100; i++) {
+                expected = LocalDate.now().dayOfMonth();
+                actual = DayOfMonth.now().value();
+                if (expected === actual) {
+                    break;
+                }
             }
-            assertEquals(LocalDate.now().dayOfMonth(), test.value());
+            assertEquals(expected, actual);
```
perceptron8, line 92: Oh, it can pass if expected and actual are undefined (after 100 unsuccessful retries) :/
I'll need to fix this once again.

packages/extra/test/reference/DayOfMonthTest.js
```diff
@@ -94,11 +99,16 @@ describe('org.threeten.extra.TestDayOfMonth', () => {
...
+            let actual;
+            for (let i = 0; i < 100; i++) {
+                expected = LocalDate.now(zone).dayOfMonth();
+                actual = DayOfMonth.now(zone).value();
+                if (expected === actual) {
+                    break;
+                }
             }
-            assertEquals(LocalDate.now(zone).dayOfMonth(), test.value());
+            assertEquals(expected, actual);
```
perceptron8, line 111: Same here.

packages/extra/test/reference/DayOfYearTest.js
```diff
@@ -81,11 +81,16 @@ describe('org.threeten.extra.TestDayOfYear', () => {
...
+            let actual;
+            for (let i = 0; i < 100; i++) {
+                expected = LocalDate.now().dayOfYear();
+                actual = DayOfYear.now().value();
+                if (expected === actual) {
+                    break;
+                }
             }
-            assertEquals(LocalDate.now().dayOfYear(), test.value());
+            assertEquals(expected, actual);
```
perceptron8, line 93: Here...

packages/extra/test/reference/DayOfYearTest.js
```diff
@@ -95,11 +100,16 @@ describe('org.threeten.extra.TestDayOfYear', () => {
...
+            let actual;
+            for (let i = 0; i < 100; i++) {
+                expected = LocalDate.now(zone).dayOfYear();
+                actual = DayOfYear.now(zone).value();
+                if (expected === actual) {
+                    break;
+                }
             }
-            assertEquals(LocalDate.now(zone).dayOfYear(), test.value());
+            assertEquals(expected, actual);
```
perceptron8, line 112: ... and here.

packages/extra/test/reference/DayOfMonthTest.js
```diff
@@ -0,0 +1,600 @@
...
+        }
+    }
+
+    TestingField.INSTANCE = new TestingField();
+
+    //-----------------------------------------------------------------------
+    // now()
+    //-----------------------------------------------------------------------
+    describe('now()', function(/*this*/) {
+        this.retries(100);
```
pithu, line 82: Yes, that's much better than a for loop 👍 . 
An alternative would be to use sinon fakeTimers, might be a bit better readable. 
BTW wouldn't be a retry of one enough? If there is the rare case of a day change between the two calls of now(), the next retry can not have a day change.

perceptron8: > An alternative would be to use sinon fakeTimers, might be a bit better readable.

In java version, you would have to mock `Clock`, which doesn't seem to be a good idea (junit itself can be using `Clock` in reports and so on). As this are reference tests, I try to make them as similar to the original as possible - they are going to be easier to maintain.

> BTW wouldn't be a retry of one enough? If there is the rare case of a day change between the two calls of now(), the next retry can not have a day change.

Next retry _should_ not have a day change, but you can't tell for sure. It's clear that more than 100 instead of 2 won't harm, but at the same time I agree that 2 could be more self-descritive. In threeten-extra there is TestOffsetDate.java, where number of retries is 100. I assumed that 100 should be applied consistently in the entire project, but we can wait to see if https://github.com/ThreeTen/threeten-extra/pull/218/commits/41cb65f04d075cba650ae8a01fa96a0360c5033c is going to be merged with same or different number of retries in DayOfMonth tests (if it's going to be merged at all).

pithu: Indeed, in java it's not recommended to mock time, but in javascript its unproblematic. Anyway, the retry solution is good to go and its a good argument to stick closer to the threeten impl. Nevertheless we should not copy bugs, etc from threeten

>  It's clear that more than 100 instead of 2 won't harm

The problem with 100 is, that in case of an other error, the test get repeated 100 times. Not a real problem, but should only be done if really required.

IMHO `this.retries(1)` is correct and therefore easier to understand. If the first test fails because of a day change, the second attempt will not fail, because the new day will have already started.

packages/extra/test/reference/DayOfMonthTest.js
pithu: good catch

packages/extra/test/reference/DayOfMonthTest.js
perceptron8: > IMHO `this.retries(1)` is correct and therefore easier to understand. If the first test fails because of a day change, the second attempt will not fail, because the new day will have already started.

It's still theoretically possible to encounter another day change after first fail (i.e. after NTP sync), so `this.retries(1)` isn't necessarily 100% correct, but I agree that 1 retry may be easier to understand. After all, it's still theoretically possible to encounter 100 fails in a row even if now() was implemented correctly.

I'll change the number of retries to 1 when I get back home.

perceptron8: Done. I had to force push because of reword, let me be excused.

pithu (commented):
Thanks a lot, looks good to me now. Before merging, i want to do a release first of the current state. Will do it in the next days.

pithu (approved):

PR #618, opened 2022-08-14
save CPU time by reducing memory pressure caused by `expect`

### LINKED ISSUE #617: extra package tests are failing
extra package tests are failing on IE 11.0 (Windows 10) two times -> `org.threeten.extra.TestDayOfYear compareTo() test_compareTo FAILED` 

see https://github.com/js-joda/js-joda/runs/7825986595?check_suite_focus=true#step:8:237

@perceptron8 could you have a look ? Is it possible that there is some javascript code that is not compatible with ie11 ? either in the `test` or `src` code. 

We would need to fix this before the next release. We promised our users to be ie11 compatible.

perceptron8:
fixes #617


pithu (approved):
LGTM, tests should be fast enough now

PR #621, opened 2022-08-18
fix rounding in OffsetTime and YearMonth

perceptron8:
Currently `OffsetTime.until` uses floor rounding and `YearMonth.until` doesn't use rounding at all. In second case, even the tests are broken! It's clear that integer division = rounding towards 0 should be used. And that is precisely what this PR fixes.


pithu (approved):
LGTM

perceptron8:
@pithu Please let me know if there is anything that prevents this fix from being applied.

PR #626, opened 2022-09-01
Upgrade tzdb to 2022c

pithu:
update tzdb and introduce a new reduced range. 2012-2022 is almost out of scope, time to setup a new minimal range +- 5 years


InExtremaRes (approved):
I've tested it locally and everything seems ok!

PR #630, opened 2022-09-23
lengthInDays(): POSITIVE_INFINITY instead of NaN if range is unbounded

### LINKED ISSUE #629: extra package tests are failing on windows ie11
extra package tests are failing on windows ie11, @perceptron8 could you have a look ? Saucelab tests have a problem with Number.isNan

perceptron8:
In case of unbounded ranges, `LocalDateRange.lengthInDays()` should return `Number.POSITIVE_INFINITY` instead of `Number.NaN`.

This also fixes #629.

### DIFF
--- a/packages/core/test/reference/LocalDateTest.js
+++ b/packages/core/test/reference/LocalDateTest.js
@@ -351,7 +351,7 @@ describe('org.threeten.bp.TestLocalDate', () => {
 
         it('factory_ofYearDay_ints_yearTooLow', () => {
             expect(() => {
-                LocalDate.ofYearDay(Number.MIN_SAFE_INTEGER, 1);
+                LocalDate.ofYearDay(MathUtil.MIN_SAFE_INTEGER, 1);
             }).to.throw(DateTimeException);
         });
     });
--- a/packages/extra/src/LocalDateRange.js
+++ b/packages/extra/src/LocalDateRange.js
@@ -670,13 +670,13 @@ export class LocalDateRange {
      * Obtains the length of this range in days.
      * 
      * This returns the number of days between the start and end dates.
-     * Unbounded ranges return `Number.NaN`.
+     * Unbounded ranges return `Number.POSITIVE_INFINITY`.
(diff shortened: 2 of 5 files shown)


perceptron8:
@pithu Please take a look.

pithu:
Looks like ie11 has an issue here as well https://github.com/js-joda/js-joda/blob/3d5f2093e029db2708bd006b814802685adc4a34/packages/extra/test/reference/OffsetDateTest.js#L354.

Looks like Number.MIN_SAFE_INTEGER  is not supported ? And what about Number.POSITIVE_INFINITY are you sure it exists in ie11. Unfortunately mdn is not listing ie11 anymore ....

pithu:
see here https://github.com/js-joda/js-joda/actions/runs/3108748813/jobs/5038283718#step:8:288

and whats happening here  https://github.com/js-joda/js-joda/actions/runs/3108748813/jobs/5038283718#step:8:300 ?

perceptron8:
> Looks like Number.MIN_SAFE_INTEGER is not supported ? And what about Number.POSITIVE_INFINITY are you sure it exists in ie11.

Looks like it's supported:
https://caniuse.com/?search=Number.POSITIVE_INFINITY

As for the rest of issues, I'll get VM with IE11 from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ and run karma by myself.

perceptron8:
It's all green on IE11.

pithu:
Cool, let's proof it by CI. 

BTW, since mdn dropped ie, we should think about to remove ie11 support

pithu:
Hmm timeout for ie11 -> https://github.com/js-joda/js-joda/actions/runs/3114145897/jobs/5049644795#step:8:299
let's rerun

perceptron8:
Timeout once again.

perceptron8:
Fix is going to be more or less similar to #617.

perceptron8:
See #631 please.

PR #631, opened 2022-09-23
save CPU time by reducing memory pressure caused by expect (again)

perceptron8:


perceptron8:
@pithu May I gently ask you for slight rewrite of my last 3 commits? I've mistakenly used my default git settings and am a little concerned about privacy ("author" of the commits should be the same as "committer"). I know that this would require you to force push on main branch, but, well, at least I tried...?

pithu:
> @pithu May I gently ask you for slight rewrite of my last 3 commits? I've mistakenly used my default git settings and am a little concerned about privacy ("author" of the commits should be the same as "committer"). I know that this would require you to force push on main branch, but, well, at least I tried...?

Sorry, i can't force push to main, but you can at least fix the commits of this PR

perceptron8:
Oh, that's unfortunate. If really nothing can be done, then it doesn't matter if there is 1 or 100 such commits, so feel free to merge.

pithu (approved):

PR #636, opened 2022-10-14
fix urls in timezone package

Inok:
Now "homepage" and "bugs url" in `timezone` package leads to the current monorepo instead of the old repo.


pithu:
Thanks for contributing

PR #637, opened 2022-10-14
update urls in locale packages

Inok:
Updated "homepage", "bugs url" and "repository url" to lead to the new monorepo instead of the old repo.


pithu (commented):
Thanks for contributing. This files are generated, the template has to be changed as well https://github.com/packages/locale/utils/create_packages.js

Inok:
Sorry, fixed the template as well. Thank you!

PR #652, opened 2022-12-06
allow a union in Interval of

xenoterracide:
this is needed for code like this to compile

```ts
    const end = offeredEnd
      ? Instant.ofEpochMilli(offeredEnd?.getUTCMilliseconds())
      : Duration.ofSeconds(Number.MAX_VALUE);

    const start = Instant.ofEpochMilli(offeredStart.getUTCMilliseconds());
    this._offeredDuring = Interval.of(start, end);
```

### DIFF
--- a/packages/extra/test/typescript_definitions/js-joda-extra-tests.ts
+++ b/packages/extra/test/typescript_definitions/js-joda-extra-tests.ts
@@ -86,14 +86,16 @@ function test_DayOfYear() {
 }
 
 function test_Interval() {
+    Interval.parse("2019-08-09T20:38:43.298Z/2019-08-09T20:38:43.298Z");
+
     const instant = Instant.now();
     const duration = Duration.ofHours(1);
+    const endExclusiveOrDuration: Instant | Duration = duration;
 
     Interval.of(instant, duration);
-
-    Interval.parse('2019-08-09T20:38:43.298Z/2019-08-09T20:38:43.298Z');
-
+    Interval.of(instant, endExclusiveOrDuration);
     const interval = Interval.of(instant, instant);
+    
     interval.start();
     interval.end();
     interval.isEmpty();
--- a/packages/extra/typings/js-joda-extra.d.ts
+++ b/packages/extra/typings/js-joda-extra.d.ts
@@ -46,8 +46,7 @@ export class DayOfYear extends TemporalAccessor {
 }
 
 export class Interval {
-    static of(startInclusive: Instant, endExclusive: Instant): Interval;
-    static of(startInclusive: Instant, duration: Duration): Interval;
+    static of(startInclusive: Instant, endExclusiveInstantOrDuration: Instant | Duration): Interval;
     static parse(text: string): Interval;
 
     private constructor(startInclusive: Instant, endExclusive: Instant);


perceptron8:
Seems to be caused by https://github.com/microsoft/TypeScript/issues/14107.

Another, slightly simpler example of code that doesn't compile:
```typescript
const start = Instant.EPOCH;
const endExclusiveOrDuration = Math.random() < 0.5 ? Instant.EPOCH : Duration.ZERO;
const interval = Interval.of(start, endExclusiveOrDuration);
```

InExtremaRes:
Hi and thank you for contributing. I have a couple of comments:

1. The change you are proposing makes the overload definition redundant. I think you can remove the other two and keep just the one with the union type.
2. I would suggest adding some tests. Make sure to test a simplified version of the code you posted and see if the rest of the tests for this method make sense to you, so we don't end up having a breaking change.
3. The rest of the code tends to use names as `instantOrDuration` for parameters with union types. For consistency you can maybe change this one to `endExclusiveOrDuration` or something similar.

xenoterracide:
> The change you are proposing makes the overload definition redundant. I think you can remove the other two and keep just the one with the union type.

yeah I thought of this. wasn't sure how you'd want to deal with it. Basically I did as little as possible

> I would suggest adding some tests. Make sure to test a simplified version of the code you posted and see if the rest of the tests for this method make sense to you, so we don't end up having a breaking change.

do you have typescript tests? I didn't see them

> The rest of the code tends to use names as instantOrDuration for parameters with union types. For consistency you can maybe change this one to endExclusiveOrDuration or something similar.

isn't that what I called it? I think you typo-ed your suggestion. I thinking you want `endInstantExclusiveOrDuration`

InExtremaRes:
> > The change you are proposing makes the overload definition redundant. I think you can remove the other two and keep just the one with the union type.
> 
> yeah I thought of this. wasn't sure how you'd want to deal with it. Basically I did as little as possible

I see. I think deleting the other two is very likely safe, but better ensure that with some tests.

> do you have typescript tests? I didn't see them

[Here](/packages/extra/test/typescript_definitions/js-joda-extra-tests.ts). The file is just type-checked, it never runs.

> isn't that what I called it? I think you typo-ed your suggestion. I thinking you want `endInstantExclusiveOrDuration`

I think you called it just `end`, but your suggestion is good too, whichever makes more sense to you and is consistent with `core` types.

pithu:
Thanks @InExtremaRes for jumping in !

xenoterracide:
https://github.com/packages/extra/test/typescript_definitions/js-joda-extra-tests.ts 404's ;)

InExtremaRes:
@xenoterracide I must've misscopied the filepath, but the file is easy to find inside the project.

xenoterracide:
sorry it took so long, busy little bee... this work?

packages/extra/test/typescript_definitions/js-joda-extra-tests.ts
```diff
@@ -88,8 +88,10 @@ function test_DayOfYear() {
 function test_Interval() {
     const instant = Instant.now();
     const duration = Duration.ofHours(1);
+    const endExclusiveOrDuration: Instant | Duration = duration;
 
     Interval.of(instant, duration);
+    Interval.of(instant, endExclusiveOrDuration);
```
pithu, line 94: What about testing `Interval.of(instant, endExclusiveInstant);` as a third case ?

pithu (commented):
Beside the one comment, LGTM.
@InExtremaRes do you have any further objections ?

packages/extra/test/typescript_definitions/js-joda-extra-tests.ts
xenoterracide: that's already in here, it's just not showing in the diff. https://github.com/js-joda/js-joda/pull/652/files#diff-1c99bd2ff528e18c76d17c2f5f739141de1f996ddd78696f35ac965c3390a1d4R98

pithu: You are testing two times with the same instance `duration`, what i miss here is a test with a value of class Instant, eg  `Interval.of(instant, instant)`

xenoterracide: yes, because it's on a different line, it was already there. For some reason it's way down. Do you want me to move it up? Trying to make as few changes as possible.

pithu: Ah ok, there it is, yes maybe we should move `Instant.parse` before the `Instant.of` tests

pithu (approved):

InExtremaRes:
Hi! I'm currently out of city and without any computer to really test it right now. I'll be back in a couple of days. However, if this PR is urgent, I fell very confident in the change it proposes and the tests can (if needed) improved later.

xenoterracide:
ok, I moved things around, better?

xenoterracide:
p.s. it's not urgent. I think I worked around it, or realized something else needed doing anyways. Sorry that it took yet another 2 days. I've come down with covid, yay!

pithu:
@xenoterracide thanks for contributing and sorry for the delay

pithu:
And get well soon

PR #657, opened 2022-12-12
add extra/Temporals

perceptron8:
I was looking for `Temporals` today (`nextWorkingDayOrSame()` in particular), but found nothing. Here it is then.

Implemented:
- nextWorkingDay()
- nextWorkingDayOrSame()
- previousWorkingDay()
- previousWorkingDayOrSame()
- parseFirstMatching(text, query, ...formatters)
- convertAmount(amount, fromUnit, toUnit)

Missing:
- chronoUnit(timeUnit)
- timeUnit(chronoUnit)
- duration{From,To}Double
- multiply

`chronoUnit` and `timeUnit` were ommitted as there is no `TimeUnit`, thus no need for conversion.

`duration{From,To}Double` and `multiply` are ommitted until someone asks for them. Without `BigDecimal`, they are not so easy to implement.

### DIFF
--- a/packages/core/src/temporal/IsoFields.js
+++ b/packages/core/src/temporal/IsoFields.js
@@ -812,7 +812,7 @@ class Unit extends TemporalUnit {
     }
 
     toString() {
-        return name;
+        return this._name;
     }
 }
 
--- a/packages/extra/src/Temporals.js
+++ b/packages/extra/src/Temporals.js
@@ -1,5 +1,5 @@
 /**
- * @copyright (c) 2016, Philipp Thürwächter & Pattrick Hüper & Michał Sobkiewicz
+ * @copyright (c) 2015-present, Philipp Thürwächter & Pattrick Hüper & Michał Sobkiewicz
  * @copyright (c) 2007-present, Stephen Colebourne & Michael Nascimento Santos
  * @license BSD-3-Clause (see LICENSE in the root directory of this source tree)
  */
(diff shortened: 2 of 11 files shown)


perceptron8:
Some assertions were in wrong order, I had to fix them.

packages/core/src/temporal/IsoFields.js
```diff
@@ -812,7 +812,7 @@ class Unit extends TemporalUnit {
     }
 
     toString() {
-        return name;
+        return this._name;
```
pithu, line 815: nice bugfix :+1: What about a test for it ?

packages/extra/src/Temporals.js
```diff
@@ -0,0 +1,296 @@
+/**
+ * @copyright (c) 2016, Philipp Thürwächter & Pattrick Hüper & Michał Sobkiewicz
```
pithu, line 2: In this case, since you are the main contributor to the extra package and creator of this file, i think its fine to add your name, but what about to change the date to `2015-present` ?

pithu (commented):
Sorry agin for the late reply and thanks for contributing.

Great work ! LGTM, i left some minor comments

packages/core/src/temporal/IsoFields.js
perceptron8: Can do! ;)

packages/extra/src/Temporals.js
perceptron8: Sure, I'll change the date.

perceptron8:
> Sorry agin for the late reply and thanks for contributing.
> 
> Great work ! LGTM, i left some minor comments

Thank you for your comments!
This PR is ready to be reviewed again.

pithu (approved):
LGTM

PR #658, opened 2022-12-21
Fix sauce

pithu:
Fix saucelabs tests


pithu:
Will be fixed with #659

PR #663, opened 2023-01-01
support more temporals in `convert()`

perceptron8:
This PR rewrites `convert()`, so it can properly support temporals like `OffsetDateTime`, `OffsetDate` and others.


perceptron8:
I know that `??` (nullish coalescing operator) should be used instead of `||` (or) in the following fragment, but it would require changing `.eslintrc.js` / `parserOptions.ecmaVersion` to `2020`+ which I was afraid to do without approval.

```javascript
const localTime = temporal.query(TemporalQueries.localTime()) || LocalTime.MIDNIGHT;
const zoneId = temporal.query(TemporalQueries.zone()) || zone || ZoneId.systemDefault();
```

PR #664, opened 2023-01-01
introduce Date$

perceptron8:
This PR introduces `Date$` class with two static methods:
- `Date$.from(Instant): Date`,
- `Date$.toInstant(Date): Instant`.

They are meant as a replacement for `nativeJs()` and `convert()` which probably should be deprecated. It's very easy to get `Instant` from temporals like `ZoneDateTime` and `OffsetDateTime` and easy enough from `LocalDate`.

I believe it's more beneficial to make clear that `Date` (as a timestamp) corresponds almost 1:1 to `Instant` (marks point on the timeline with similar precision and no notion of time zone) than to pretend, that temporals like `LocalDate` have unambiguous representation as a `Date` (even when time-zone is provided).

Naming is based on:
- https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#from-java.time.Instant-
- https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#toInstant--

If one needs momentjs support, this naming convention suggests `moment$.from(Instant): Moment` and `moment$.toInstant(Moment): Date`, which is - i hope - clear and consistent.

If this PR makes it's way to main, documentation needs to be updated (deprecation of `nativeJs()` and `convert()` + rewrite of convert-native.md).

As always, I'm open for discussion.

Happy New Year!

### DIFF
--- a/packages/core/src/Date_.js
+++ b/packages/core/src/Date_.js
@@ -1,0 +1,37 @@
+/*
+ * @copyright (c) 2015-present, Philipp Thürwächter, Pattrick Hüper & js-joda contributors
+ * @license BSD-3-Clause (see LICENSE in the root directory of this source tree)
+ */
+
+import { requireInstance, requireNonNull } from './assert';
+import { Instant } from './js-joda';
+
+export class Date$ {
+    /**
+     * Convert an `Instant` object to a `Date`.
+     * @param {!Instant} instant - an `Instant` object
+     * @returns {Date} a `Date` object corresponding to given `instant`
+     */
+    static from(instant) {
+        requireNonNull(instant, 'instant');
+        requireInstance(instant, Instant, 'instant');
+        return new Date(instant.toEpochMilli());
+    }
+
+    /**
+     * Converts a `Date` object to an `Instant`.
+     * @param {!Date} date - a `Date`
+     * @returns {Instant} - an `Instant` corresponding to given `date`
+     */
+    static toInstant(date) {
+        requireNonNull(date, 'date');
+        requireInstance(date, Date, 'date');
+        return Instant.ofEpochMilli(date.getTime());
+    }
+
+    /**
+     * @private
+     */
+    constructor() {
+    }
+}
--- a/packages/core/src/js-joda.js
+++ b/packages/core/src/js-joda.js
@@ -78,6 +78,8 @@ import * as assert from './assert';
 
 import { convert } from './convert';
 import { nativeJs } from './nativeJs';
+import { Date$ } from './Date_';
+
 import { bindUse } from './use';
 
 const _ = {
@@ -94,6 +96,7 @@ const jsJodaExports = {
     _,
     convert,
     nativeJs,
+    Date$,
     ArithmeticException,
     DateTimeException,
     DateTimeParseException,
@@ -162,6 +165,7 @@ export {
     use,
     convert,
     nativeJs,
+    Date$,
     ArithmeticException,
     DateTimeException,
     DateTimeParseException,
--- a/packages/core/test/Date_Test.js
+++ b/packages/core/test/Date_Test.js
@@ -1,0 +1,50 @@
+/*
+ * @copyright (c) 2015-present, Philipp Thürwächter, Pattrick Hüper & js-joda contributors
+ * @license BSD-3-Clause (see LICENSE in the root directory of this source tree)
+ */
+
+import { expect } from 'chai';
+
+import './_init';
+
+import { Instant } from '../src/Instant';
+import { Date$ } from '../src/Date_';
+import { IllegalArgumentException, NullPointerException } from '../src/errors';
+
+describe('Date$', () => {
+    it('.from(Instant) should convert Instant to Date', () => {
+        for (const epochMilli of [0, 1000, 60 * 1000, 60 * 1000, 24 * 60 * 60 * 1000]) {
+            const instant = Instant.ofEpochMilli(epochMilli);
+            const date = Date$.from(instant);
+            expect(date).to.be.instanceOf(Date);
+            expect(date.getTime()).to.equal(epochMilli);
+        }
+    });
+
+    it('.from(Instant) should throw IAE when called with !Instant', () => {
+        expect(() => Date$.from(0)).to.throw(IllegalArgumentException);
+    });
+
+    it('.from(Instant) should throw NPE when called with null or undefined', () => {
+        expect(() => Date$.from(null)).to.throw(NullPointerException);
+        expect(() => Date$.from(undefined)).to.throw(NullPointerException);
+    });
+
+    it('.toInstant(Date) should convert Date to Instant', () => {
+        for (const epochMilli of [0, 1000, 60 * 1000, 60 * 1000, 24 * 60 * 60 * 1000]) {
+            const date = new Date(epochMilli);
+            const instant = Date$.toInstant(date);
+            expect(instant).to.be.instanceOf(Instant);
+            expect(instant.toEpochMilli()).to.equal(epochMilli);
+        }
+    });
+
+    it('.toInstant(Date) should throw IAE when called with !Date', () => {
+        expect(() => Date$.toInstant(0)).to.throw(IllegalArgumentException);
+    });
+
+    it('.toInstant(Date) should throw NPE when called with null or undefined', () => {
+        expect(() => Date$.toInstant(null)).to.throw(NullPointerException);
+        expect(() => Date$.toInstant(undefined)).to.throw(NullPointerException);
+    });
+});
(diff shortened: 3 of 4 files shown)


pithu:
Hi @perceptron8 

i wish you a happy new year as well and thanks for the new PR's.

> I believe it's more beneficial to make clear that Date (as a timestamp) corresponds almost 1:1 to Instant (marks point on the timeline with similar precision and no notion of time zone) than to pretend, that temporals like LocalDate have unambiguous representation as a Date (even when time-zone is provided).

Agree 100% !

But actually i am not convinced, that we should add a new class called Date similar to the jdk. The Date class was one of the biggest failures in early jdk development, and that js used that class as a template in the early days of javascript development started all the trouble with bad Date design. Using a class called `Date` can produce a lot of confusion: Is it a jdk Date or a Javascript Date ? Why it just have to static functions, etc. And why we should add this new class, just for add some very simple convenience functionality. 

IMO we just need two convenient methods, `nativeToInstant(date:Date | moment): Instant` and `instantToNative(instant:Instant): Date` and mark the two existing "converters" `nativeJs(date, zone)` and `convert(temporal)` as deprecated. 

Concerning your other PR, I don't think its a good idea to add functionality to convert other temporals than an Instant to make it more clear, that an Instant is the closest representation of a js Date. Furthermore the functionality could only be used  to convert to js Date. I would prefer to add some examples to the documentation (mainly all one liners) instead

perceptron8:
> But actually i am not convinced, that we should add a new class called Date similar to the jdk. The Date class was one of the biggest failures in early jdk development, and that js used that class as a template in the early days of javascript development started all the trouble with bad Date design. Using a class called `Date` can produce a lot of confusion: Is it a jdk Date or a Javascript Date ?

I'm confused. Why would one think that this class have anything in common with jdk Date? You don't annotate every `Date` usage within js-joda with explanation that it's not jdk Date, don't you? It's perfectly clear IMO.

> Why it just have to static functions, etc. And why we should add this new class, just for add some very simple convenience functionality.

`Date$` acts here as some kind of namespace. It's not `Date`, so there should be no confusion. It could be `Dates` too (guava convention), but I think `Date$.from` is more readable despite $ (it reads more or less like: "date from xyz"). If `Date$` is too much for you, I once again propose `dateFromInstant` and `dateToInstant` based on java naming.

> IMO we just need two convenient methods, `nativeToInstant(date:Date | moment): Instant` and `instantToNative(instant:Instant): Date` and mark the two existing "converters" `nativeJs(date, zone)` and `convert(temporal)` as deprecated.

I'm not convinced about "native" part of names you propose. It appears that there will be another, more mature "native" date-related set of classes available in js runtime very (?) soon (https://tc39.es/proposal-temporal/docs/).

I'm also not convinced about supporting momentjs - which is "a legacy project in maintenance mode" (https://momentjs.com/docs/#/-project-status/) - in new code (backward compatibility is another story).

> Concerning your other PR, I don't think its a good idea to add functionality to convert other temporals than an Instant to make it more clear, that an Instant is the closest representation of a js Date. Furthermore the functionality could only be used to convert to js Date. I would prefer to add some examples to the documentation (mainly all one liners) instead

As for the other PR, it's definitely _not_ about adding functionality (unfortunate support for `LocalDate` and `zone` parameter was here long time before me). It's about fixing "incomplete" implementation. If `convert` can create `Date` from `ZoneDateTime` then I can see no reason why shouldnt' it be able to create `Date` from `OffsetDateTime`. But then I'd like to deprecate it anyway.

If `convert` wasn't used before, I'd rather propose

```javascript
convert(instant).toDate();
convert(date).toInstant();
```

instead and deprecate `nativeJs`, but sadly it's too late for this kind of cleanup.

pithu:
`Date` is ambiguous, already used in jdk and javascript and its a very general term. I just remember the long discussions for the temporal proposal. Looks like naming is the most challenging task here. 

`convert` is very general as well and the API can lead to strange code like `convert(instant).toInstant()`.

If i don't oversee something, we have exactly two use cases. And both use cases are only for a little bit of syntactic sugar for convenience. And i think we just need two functions for the two use cases with "good" names.

1. convert a date or moment (or any other Object that is returning the `epochMillis` with valueOf) to an Instant
2. convert an Instant to a javascript Date instance

I agree that the term `nativeJs` in `nativeJsToInstant` or `instantToNativeJs` is not clear as well, can we find a better  naming ?

And we should update the documentation with some examples.

WDYT ?

perceptron8:
In whole js-joda, `Date` means js date. But ok, `Date$` is a little awkward name. Let's just find 2 names and forget about "namespacing". I would stick to `dateFromInstant` and `dateToInstant` (and I mean js Date, not jdk Date! ;) ). Why 'dateFromInstant` and not `instantToDate`? For the same reason by which `Instant` doesn't have `toDate` in jdk - to avoid polluting new domain with references to old domain classes. I think good decision was made here and we can rely on it. We can't (and shouldn't) monkey patch js `Date` with `fromInstant` / `toInstant`, that's why i propose `dateFromInstant` / `dateToInstant`.

BTW `convert(instant)` could return instance of class having only `.toDate()`. Similarly, `convert(date)` could return instance of class having only `toInstant()`. So "strange code" (`convert(instant).toInstant()`) wouldn't be a problem. The problem is, that the best name for a function (convert) is already taken, and that this function takes 2 parameters instead of 1, and that anything other than `Date` and `Instant` is being involved.

Documentation + examples are a must have, sure.

pithu:
Let's collect some naming proposals. 

One function that returns an Instant and takes as input a js `Date` or any other object where `valueOf` returns the `epochMillis`. And one function that takes an Instant and returns a js `Date`

1. `nativeJsToInstant(date)` / `instantToNativeJs(instant))`
2. `dateToInstant(date)` / `instantToDate(instant))`
3. `dateFromInstant(instant)` / `dateToInstant(date)`
4.  ...

perceptron8:
I'm not interested in anything other than f: `Date` -> `Instant` and g: `Instant` -> `Date`. Supporting "any other object where `valueOf` returns the `epochMillis`" will make `f(true)`, `f(42)` and `f('0x1337')` legal.

perceptron8:
> Supporting "any other object where `valueOf` returns the `epochMillis`" will make `f(true)`, `f(42)` and `f('0x1337')` legal.
Not directly, but in case one would implement f as `Instant.from(date.valueOf())` without additional check checking that result of `date.valueOf()` is a number. Requiring `Date` as input solves this problem and eliminates assertion. It also introduces some symmetry in types which I find appealing.

perceptron8:
Oh, but it would make `f(Math.PI)` valid (compilable) expression. Let's agree today on this one thing at least: f - however called - will get `Date` as an input. Is that OK with you, @pithu?

pithu:
What about `moment` ? And we  could throw if the argument is not an  "int" value

perceptron8:
As for "ints":

First, `Instant.ofEpochMilli(0.123456)` works fine even if not called with "int" (i'm not saying that's good; it's an example of how ambigious function taking `number` can be). Second, not every "int" is fine, only ones between about ±8,640,000,000,000,000 (see https://tc39.es/ecma262/#sec-date-objects). I guess we can throw and throw again instead of limiting parameter to the right type?

As for "momentjs":

> I'm also not convinced about supporting momentjs - which is "a legacy project in maintenance mode" (https://momentjs.com/docs/#/-project-status/) - in new code (backward compatibility is another story).

`Moment` is not as easily detectable as `Date`. And, what's more important, it's not reliably detectable (with "reflection"-like tricks).

Apart from that, I'm sure that if anyone needs `Moment` -> `Instant` conversion, then he/she probably needs `Instant` -> `Moment` conversion also.

I believe momentjs users can be gently guided on how to:

1. extend `moment` with `toInstant` using `Moment.prototype` exposed as `moment.fn` (see [https://momentjs.com/docs/#/parsing/](https://momentjs.com/docs/#/parsing/:~:text=The%20Moment%20prototype%20is%20exposed%20through%20moment.fn.%20If%20you%20want%20to%20add%20your%20own%20functions%2C%20that%20is%20where%20you%20would%20put%20them.)) + extend `Moment` with `fromInstant` if that feels idiomatic for them or
2. implement `momentFromInstant` / `momentToInstant` (what a great names!) on their own using `dateFromInstant` / `dateToInstant` which seems more than trivial given `moment.toDate()` and `moment(Date)` being available.

If built-in momentjs support is an absolute must have, I'd insist on @js-joda/moment or similar with explicit moment dependency.

Is that finally OK with you, @pithu?

pithu:
Fyi: we have some functions in mathutils to detect int values, works quit well for JavaScript 

Js-joda has 0 production dependencies, therefore adding moment as a dependency is no option 

I am fine with keeping things simple and just support Date, makes code simple and robust.

perceptron8:
Superseded by #670.
Thanks for discussion!

PR #666, opened 2023-01-02
fix and simplify nativeJs implementation

perceptron8:
1. Switching by constructor name can break code that are using certain bundlers.

2. Both `Date` and `Moment` provide primitive conversion with `valueOf()`.
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf
- https://momentjs.com/docs/#/displaying/unix-timestamp-milliseconds/

It turns out that there is no need for any conditional constructs.

### DIFF
--- a/packages/core/src/nativeJs.js
+++ b/packages/core/src/nativeJs.js
@@ -4,7 +4,6 @@
  */
 
 import { requireNonNull } from './assert';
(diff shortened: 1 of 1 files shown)


perceptron8:
Superseded by #667.

PR #667, opened 2023-01-02
simplify nativeJs implementation

perceptron8:
Both `Date` and `Moment` provide primitive conversion with `valueOf()`.

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf
- https://momentjs.com/docs/#/displaying/unix-timestamp-milliseconds/

It turns out that there is no need for any conditional constructs.

### DIFF
--- a/packages/core/src/nativeJs.js
+++ b/packages/core/src/nativeJs.js
@@ -4,7 +4,6 @@
  */
 
 import { requireNonNull } from './assert';
-import { IllegalArgumentException } from './errors';
 import { Instant, ZoneId } from './js-joda';
 
 /**
@@ -16,10 +15,5 @@ import { Instant, ZoneId } from './js-joda';
 export function nativeJs(date, zone = ZoneId.systemDefault()) {
     requireNonNull(date, 'date');
     requireNonNull(zone, 'zone');
-    if(date instanceof Date) {
-        return Instant.ofEpochMilli(date.getTime()).atZone(zone);
-    } else if(typeof date.toDate === 'function' &&  date.toDate() instanceof Date) {
-        return Instant.ofEpochMilli(date.valueOf()).atZone(zone);
-    }
-    throw new IllegalArgumentException('date must be a javascript Date or a moment instance');
+    return Instant.ofEpochMilli(date.valueOf()).atZone(zone);
 }


packages/core/src/nativeJs.js
```diff
@@ -16,10 +15,5 @@ import { Instant, ZoneId } from './js-joda';
 export function nativeJs(date, zone = ZoneId.systemDefault()) {
     requireNonNull(date, 'date');
     requireNonNull(zone, 'zone');
-    if(date instanceof Date) {
-        return Instant.ofEpochMilli(date.getTime()).atZone(zone);
-    } else if(typeof date.toDate === 'function' &&  date.toDate() instanceof Date) {
-        return Instant.ofEpochMilli(date.valueOf()).atZone(zone);
-    }
-    throw new IllegalArgumentException('date must be a javascript Date or a moment instance');
+    return Instant.ofEpochMilli(date.valueOf()).atZone(zone);
```
pithu, line 18: nativeJs will behave different for some edge case. The previous code throws if not instance of `Date` or `Moment`. With this change it might not throw in some edge cases and break users code in some rare cases. I would prefer to not change contract here, mark the function as deprecated and find a new implementation.

I would even think about to change `date.valueOf()` back to `date.toDate().getTime()`. It would make this code better readable and make 100% the same as before.

WDYT ?

perceptron8: That's only partially true. nativeJs wasn't throwing even if date was not Moment (it was only requiring some toDate method returing Date, so one could pass something other than Moment). You had no objections here! Why invoking valueOf bothers you? The contract is the same: nativeJs supports Date and Moment instances - and that's it.

I would still stay that this change makes code simpler and more readable - and that is so obvious that I can't imagine, how that needs to be argued.

pithu: Its not about code, its about keeping the semver contract and backwards compatibility https://semver.org/. 

The contract of nativeJs for this major version is: 
For a given instance of Date or for an instance that has a function `toDate` that is returning a Date, return an instance that is implementing the TemporalAccessor interface. Your implementation is not 100% backward compatible and breaking this contract, because it would not throw an exception eg for `nativeJs(new Number(23))`. To release this change, we would have to make a major release, what makes no sense for a small change like this.

What you usually do in such cases, you deprecate this function. So that you don't break existing code, that depends on @js-joda/core.  And add new functionality (new function/ method / etc) with a reasonable name, so that we can make a minor release for that change.

Does this answer your question ?

perceptron8: I don't think that throwing an exception was indeed part of contract - esdoc doesn't mention anything about that. And I checked `nativeJs` history from the first commit. It was rather incidental implementation detail to throw (what else left to do?) - and this was slightly broken already as I showed, as it was possible to pass something other than `Moment` (by necessity, sure, as moment isn't one of js-joda dependencies). IMO contract seems rather to be: transform `Date` / `Moment` to `TemporalAccessor`. Even stronger already: to `ZonedDateTime`. And it _is_ fulfilled. Invoking `nativeJs` with anything other than `Date` or `Moment` was illegal by the contract before just as is illegal now. We should be able to safely assume that no one was using `nativeJs` for validation purposes or something - this would be ridiculous. IMO this _is_ backward compatible change.

I woudn't resist adding `assert(date instanceof Date || date instanceof Moment);` if I could refer to `Moment`. I can't for obvious reasons. Using `typeof date.toDate === 'function' &&  date.toDate() instanceof Date` is less performant, `toDate()` is called twice and so on. It's just a mess that I'd like to avoid.

As for deprecation, take https://developer.mozilla.org/en-US/docs/Web/API/structuredClone as an example. Let's say you introduce `BigInt` into runtime. Should you deprecate `structuredClone` because you want to support another type of value or not? Obviously not. And that's even not what happens to `nativeJs`, as signature stays in place.

perceptron8: Here's another example, that will more likely convince you. See
- https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/Duration.html#dividedBy(long) and
- https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/Duration.html#dividedBy(java.time.Duration).

Can you agree that the addition of `Duration.dividedBy(Duration)` in JDK 9 was backward compatible? I know that java has overloads and that above two are entirely different methods at compile time, but it doesn't change fact that similar addition in js-joda is possible in a backward compatible manner. Of course, if someone called `Duration.dividedBy(Duration)` just to get an exception (before addition of this method), then yes, it would break his code, but I can't see this as a valid use case. It should be considered a programmer's error.

On the other hand, I do understand that you would like to prevent calls like `nativeJs(true)`, `nativeJs(42)` and `nativeJs('0x1337')`. I'll try to find some concise way to check for `Date` and `Moment`. If only the latter wasn't listed as allowed...

pithu:
Again, Its not about code or documentation, it's about semver. Semver means, that a minor change is not breaking other people code. As you remember we already break other people code by just changing the detection of moment and date.

In this case it might be a bit petty because it's an undocumented edge case, but since we're going to deprecate this feature anyway, I see no reason to break other people's code if they have a caret dependency to @js-joda/core in their code.

Eg `nativeJs(new Number(23))` have been throwing an error before and should keep this behaviour. This PR would change this behaviour.

Therefore i will close the PR for now.

PR #670, opened 2023-01-11
augment Date

perceptron8:
This is my final attempt on replacing `convert` and `nativeJs` with something less error-prone and more convenient to use.

`Date` gained 2 additional methods:

```javascript
static Date.from(Instant): Date
Date.toInstant(): Instant
```

so the following became possible:

```javascript
const dateFromInstant = Date.from(Instant.now());
const instantFromDate = new Date().toInstant();

const momentFromInstant = moment(Date.from(Instant.now()));
const instantFromMoment = moment().toDate().toInstant();
```

I hope that augmenting native class won't feel too controversial for you. After all, most of polyfills do the same. It seems to be the cleanest solution considering previously explored alternatives and endless naming-related battles. I'm sure that proposed way of converting between `Date` and `Instant` will feel very natural to everyone, especially to java users.

Typings and documentation have been updated accordingly. All IDEs I've tested seem to work as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 ## Why yet another JavaScript date and time library?
 
-- Popular JavaScript date libraries like [moment](https://momentjs.com/) or [date-utils](https://github.com/continuouscalendar/dateutils) are **wrappers** around the native JavaScript `Date` object, providing syntactic sugar. The native `Date` object always consist of a date, time and a timezone part. In contrast, js-joda is a **standalone** date and time implementation.
+- Popular JavaScript date libraries like [moment](https://momentjs.com/) or [date-utils](https://github.com/continuouscalendar/dateutils) are **wrappers** around the native JavaScript `Date` object, providing syntactic sugar. In contrast, js-joda is a **standalone** date and time implementation.
 
 - The API has a **domain-driven design** with classes for each of the different use cases, like `LocalDate`, `ZonedDateTime` or `Period`. For example, `LocalDate` allows you to handle dates without times (like birthdays or holidays) in a clean and error-safe way, especially if these dates are persisted to an external server.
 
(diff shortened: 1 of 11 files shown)


perceptron8:
Hi @pithu,

I think you should use some real arguments instead of just calling something a bad practice.

> monkey patching native classes is bad

Why? I mean in this particular case. Because, as always, you know... it depends. We're not altering existing behavior, are we? That, I agree, would be bad. So it must be about potential naming conflicts, right? Or is it about code being harder to understand for some reason?

> patching the native Date object even more

Well, IMO, it's quite contrary! Why? Because `Date`, being effectively deprecated, is much, much less likely to be "natively" extended than any other class. This makes potential naming conflicts much less probable than they would be in case of, say, extending `Array` or `String`.

As for naming and "methods vs functions" battle, see https://tc39.es/proposal-temporal/docs/cookbook.html#converting-between-temporal-types-and-legacy-date. It's likely that `toTemporalInstant()` will be added to `Date`, so `toInstant()` for joda's unscoped `Instant` seems to be more natural than anything - and it matches not only Java, but feels right in JS/TS world as it appears.

BTW I'd like to align with the temporal proposal even more and remove `Date.from(instant)` in favor of `new Date(instant.toEpochMilli())`. Only the documentation would have to be extended in this case ;)

I hope you'll think it all over again and come to different conclusions.

pithu:
js-joda claims to be robust, predictable and immutable. The Problem with monkey patching is, that its not predictable. 

First problem: Anyone else could monkey patch the same function. 

The other really ugly problem is, that its not guaranteed, that at a certain time an Object is already monkey patched or not. It depends on, when the  jit compiler is loading a file. And that can be different for any environment, eg depends on bundler, etc. That can lead to really bad issues hard to debug.

On the other its super easy and without any drawbacks to implement the behaviour without monkey patching. All we need are two functions / convenient helpers:
```
dateToInstant(date:Date): Instant 
dateFromInstant(instant:Instant): Date 
```

PR #674, opened 2023-01-30
Fix optional parameter type of appendInstant method

alisabzevari:
`DateTimeFormatterBuilder.appendInstant` parameter should be optional according to [code](https://github.com/alisabzevari/js-joda/blob/b23a9e2c16bc5d632f6b76b14df5983cfd7ddbfa/packages/core/src/format/DateTimeFormatterBuilder.js#L547) and [documentation](https://js-joda.github.io/js-joda/class/packages/core/src/format/DateTimeFormatterBuilder.js~DateTimeFormatterBuilder.html#instance-method-appendInstant).
I have just made the parameter optional in typing.


pithu (approved):
Thanks for fixing

Issue #600, opened 2022-03-18
Pre-built locale packages for pl, el, da


mstawick:
Hi, can you please add pre-build packages for the following locales:
* Polish - pl
* Danish - da
* Greek - el

I've created a pull request with the necessary changes (if I understood the process correctly):
https://github.com/js-joda/js-joda/pull/599

pithu:
merged #599

PR #677, opened 2023-03-20
Add pre-built packages for fi, fi-FI locales

fo-fo:
These were copied and modified from the corresponding sv, sv-SE locales.

Let me know if there's something else that needs to be done. I based these changes on https://github.com/js-joda/js-joda/issues/600 (but also updated the documentation).


pithu:
Hei @fo-fo thanks for adding fi-fi, LGTM, will merge it after the build is green

PR #692, opened 2023-08-13
add esm modules in package.json files

### LINKED ISSUE #595: @js-joda/locale_en does not have "module" in package.json
It looks like the ESM build is present but the package.json does not have the `module` section.

`"module": "dist/index.esm.js",`

alisabzevari:
to fix #595

I just added the modules in package.json files. Let me know if more work is needed.

### DIFF
--- a/packages/extra/package.json
+++ b/packages/extra/package.json
@@ -25,6 +25,7 @@
     "README.md"
   ],
   "main": "dist/js-joda-extra.js",
+  "module": "dist/js-joda-extra.esm.js",
   "scripts": {
     "prepublishOnly": "npm run build-dist",
     "test": "NODE_ENV=test npx mocha --timeout 5000 --require @babel/register ./test/*Test.js ./test/**/*Test.js ./test/**/**/*Test.js test/pluginTest_mochaOnly.js",
--- a/packages/locale/package.json
+++ b/packages/locale/package.json
@@ -17,6 +17,7 @@
     "url": "https://github.com/js-joda/js-joda/issues"
   },
   "main": "dist/js-joda-locale.js",
+  "module": "dist/js-joda-locale.esm.js",
   "typings": "typings/js-joda-locale.d.ts",
   "files": [
     "dist",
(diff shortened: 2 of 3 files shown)


pithu:
Hi @alisabzevari, thanks for contributing. For my understanding: What is the expected value of this PR and how can i test it ?

cercatrice:
Hello @pithu, is there a reason this PR was never approved? We require the same fix in order to be able to pull in `@js-joda/locale_en-us`, and `@js-joda/timezone as ES Module`s. 

Please see documentation here: https://esbuild.github.io/api/#main-fields

The package.json file for the core package has both `main` and `module` properties defined. The `module` property allows the core package to be imported as an ES Module. A similar property needs to be added to the other packages. As far as I can see, the solution proposed by @alisabzevari seems correct.

pithu:
Hi @cercatrice ,

is there a way to test it ?

alisabzevari:
Hi @pithu, 
No unfortunately, I couldn't test it. After I saw your comment, I wasn't sure this is as easy as I thought!

cercatrice:
@pithu I just reproduced via a stackblitz: https://stackblitz.com/edit/angular-webcontainer-template-cfdks6?file=src%2Fapp%2Fapp.component.ts

The warnings in the terminal are what tells you that the module is being pulled in as a CommonJS package instead of ES Modules. 
With updated imports that contain @alisabzevari's change, the warnings will go away.  

Let me know if you have any questions.

pithu:
Ok i can create a release, actually i am quite confident that it will work, but please test it, and let me know if it s working for you

pithu (approved):

cercatrice:
@pithu, Thank you for building. 
I just tested and the change works for the timezone module - so that's great! 

For the locale modules, I am testing with "@js-joda/locale_en-us@4.11.0", and unfortunately, it does not work, i.e. I am still seeing the warning that package is being imported as a CommonJS package. 

Each of these child locale packages has a package.json that probably also needs a module property - but it is unclear what the value should be.

pithu:
Looks like the esm bundles are generated in the dist folders of the specific locales. Might be possible to fix by adding it to the template for the locale package.json

cercatrice:
I thought the same, and have a branch (from Friday) to push up, but I am having trouble with auth, github won't let me push.

cercatrice:
@pithu here's the PR to get the locale modules updated to support ES modules: https://github.com/js-joda/js-joda/pull/722

PR #697, opened 2023-09-12
fix broken link

takashima0411:
Fix broken links on documentations.
But, I'm not sure fixed link is correct.


pithu (approved):
@takashima0411 thanks for fixing. The files in docs are generated, so the change there is not required but do not harm.

PR #698, opened 2023-09-16
Feature DateTimeFormatterBuilder.parseDefaulting()

m-jung:
This PR ports the `DateTimeFormatterBuilder.parseDefaulting()` Method from ThreeTen/threetenbp to js-joda in order to increase usefulness of parsers with optional fields.
ThreeTen did not include tests, so i implementend a few simple tests seperatly.
`DefaultingParser` is located in _DateTimeFormatterBuilder.js_ since it is of no use outside of it.
I've edited _js-joda.d.ts_ and _js-joda.flow.js_ because the changes seem trivial and i've got no tooling to generate them.

### DIFF
--- a/packages/core/src/format/DateTimeFormatterBuilder.js
+++ b/packages/core/src/format/DateTimeFormatterBuilder.js
@@ -175,6 +175,39 @@ export class DateTimeFormatterBuilder {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


pithu (approved):
cool stuff! thx for contributing

PR #699, opened 2023-10-06
Add more built-in date-time formats to typings

adamschoenemann:
These seem to be missing in the typings.
I left out RFC_1123_DATE_TIME since it requires also importing `@js-joda/locales`


pithu (approved):
LGTM, thanks for contributing

Issue #712, opened 2024-03-08
Duration.plusDuration seems to be missing from type definitions


syxolk:
The `plusDuration` method of the `Duration` class is documented here: https://js-joda.github.io/js-joda/class/packages/core/src/Duration.js~Duration.html#instance-method-plusDuration

However, when I have the following Typescript code

```typescript
import { Duration } from "@js-joda/core";

function main() {
    const a = Duration.ofHours(10);
    const b = Duration.ofMinutes(2);
    console.log(a.plusDuration(b));
}
```

Typescript will complain with:

```
Property 'plusDuration' does not exist on type 'Duration'.
```

My workaround is to use this:

```typescript
a.plusMillis(b.toMillis())
```

My versions:
- `@js-joda/core`: 5.6.1
- `typescript`: 5.1.6
- Node.js: 20.5.1

pithu:
Hi @syxolk. 

It is implement, but seems to be missing in the typescript definitions here https://github.com/js-joda/js-joda/blob/main/packages/core/typings/js-joda.d.ts#L1122

As an workaround use `duration.plus()` instead and / or open a PR adding plusDuration to the typings.

I hope that helps.

PR #713, opened 2024-03-08
Add type definition for Duration#plusDuration

syxolk:
closes https://github.com/js-joda/js-joda/issues/712

Adds a missing typescript type definition für `Duration#plusDuration` that is documented here: https://js-joda.github.io/js-joda/class/packages/core/src/Duration.js~Duration.html#instance-method-plusDuration

In the flow typings, it is already available: https://github.com/js-joda/js-joda/blob/c635147087af60f35b964ba3f327715436e4ba05/packages/core/typings/js-joda.flow.js#L96


pithu (approved):

pithu:
thanks a lot @syxolk

PR #720, opened 2024-05-02
Update formatting.html

andHW:
Fix DateTimeFormatter URL.


pithu:
@andHW Thanks for reporting, your fix is in the generated docs, therefore i fixed it in the source files

PR #723, opened 2024-05-13
Fix DateTimeFormatter ISO_DATE_TIME docs example

themichaellai:
The example for `DateTimeFormatter.ISO_DATE_TIME` in the docs seems like a typo — I see this:
https://github.com/js-joda/js-joda/blob/fe7febf85e4c824605e0b98d526436411413ef21/packages/core/test/format/DateTimeFormatterTest.js#L122


pithu (approved):
thanks for contributing

PR #727, opened 2024-06-17
Added toString in YearMonth type definition

SaichandChowdary:
1) Added toString() in YearMonth type definition
2) Removed duplicate toString() in LocalTime type definition

### DIFF
--- a/packages/core/typings/js-joda.d.ts
+++ b/packages/core/typings/js-joda.d.ts
@@ -1418,7 +1418,6 @@ export class LocalTime extends Temporal implements TemporalAdjuster {
     toNanoOfDay(): number;
     toSecondOfDay(): number;
     toString(): string;
-    toString(): string;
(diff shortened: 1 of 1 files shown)


packages/core/typings/js-joda.d.ts
```diff
@@ -1600,6 +1599,7 @@ export class YearMonth extends Temporal implements TemporalAdjuster {
     withMonth(month: number): YearMonth;
     withYear(year: number): YearMonth;
     year(): number;
+    toString(): string;
```
pithu, line 1602: Could you keep lexical order, move up after toJSON()

pithu (changes requested):
Could you please keep lexical order

packages/core/typings/js-joda.d.ts
SaichandChowdary: done

pithu (approved):
cool, thanks !

PR #737, opened 2024-09-09
Improve MathUtil.verifyInt performance

pithu:


InExtremaRes:
Hi! I'm very sorry for being absent, some family issues kept me away for a while. I've tried to look at opened PRs from time to time, specially if they had changes to TS files. I hope everything starts going back to normal now.

In any case, here I'm curious about the `Number()` call. Why was it necessary? Was tests were failing without it?

pithu:
Hello @InExtremaRes, nice to hear from you ! 
Check the other PR #734 to see the failing test, its about type conversion from string. Without the Number it would be a breaking change.
I am unsure to merge this one, because it would break ie11, the failing test in this MR

InExtremaRes:
> Hello @InExtremaRes, nice to hear from you ! Check the other PR #734 to see the failing test, its about type conversion from string. Without the Number it would be a breaking change.

And that use was intended from the beginning?

> I am unsure to merge this one, because it would break ie11, the failing test in this MR.
> 
> We could add a polyfill, but it looks like this needs some package upgrades and this is the point where i stopped for now ...

Is there a limit or discussion regarding how long IE11 will be supported? I remember a few instances where a very natural improvement to the code has been discarded because of compatibility issues with IE11.

JonDum:
IE11 was officially retired over two years ago and has less then 0.16% usage on the web. If no one else will step up, I'd say it's very safe at this point to drop support for IE11

pithu:
> IE11 was officially retired over two years ago and has less then 0.16% usage on the web. If no one else will step up, I'd say it's very safe at this point to drop support for IE11

I agree, we should remove remove all references and tests for IE11. For now i added a IE11 switch to fix the issue.

Issue #743, opened 2024-12-09
Long patterns for month (MMMMM) and day (EEEEE) inconsistent with SimpleDateFormat


janoma:
The [formatting documentation](https://js-joda.github.io/js-joda/manual/formatting.html) states (emphasis mine):

> Date and time formats are **specified** by date and time pattern strings using [Java SimpleDateFormat](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) codes.

In `SimpleDateFormat`, name of day in week (`E`) has a presentation of **Text**, and month in year (`M`) has a presentation of **Number** if the pattern has less than 3 letters (`M` and `MM`), and **Text** otherwise. It also specifies:

> **Text:** For formatting, if the number of pattern letters is 4 or more, the full form is used (...)

which means that patterns like `MMMM`, `MMMMM` and `MMMMMM` should yield the same result:

```ts
import { DateTimeFormatter, LocalDate } from '@js-joda/core';
import { Locale } from '@js-joda/locale_en-us';

const date = LocalDate.of(2024, 12, 9);
const pattern4 = DateTimeFormatter.ofPattern('MMMM').withLocale(Locale.US);
const pattern5 = DateTimeFormatter.ofPattern('MMMMM').withLocale(Locale.US);
const pattern6 = DateTimeFormatter.ofPattern('MMMMMM').withLocale(Locale.US);
```

Here, all 3 patterns should result in `'December'` when doing `date.format(patternN)`. However, the actual results are
```ts
date.format(pattern4); // 'December'
date.format(pattern5); // 'D'
date.format(pattern6); // throws
```

Analogous results for name of day in week with the `EEEEE` and `EEEEEE` patterns.

Note that this *could be* justified with your own documentation, as the examples for month-of-year and day-of-week are `7; 07; Jul; July; J` and `2; Tue; Tuesday; T`, respectively. However, this is still inconsistent with the linked Java documentation.

I do not know if this is a bug in the code or in the documentation (or if maybe I just missed something else entirely). I suspect changing the behavior of the code is riskier as it would be backwards incompatible. If it's considered a documentation bug, please update it to explain the difference in behavior for these long text patterns. I can give it a go myself and create a PR for that, but I want to know if there's agreement this should be updated at all.

Using `@js-joda/core` 5.6.3 and `@js-joda/locale_en-us` 4.14.0.

pithu:
Hello @janoma, 

[...]

PR #744, opened 2024-12-11
docs: fix formatting documentation

janoma:
Explain some differences between the patterns in js-joda and `SimpleDateFormat` from java. Add some examples.

See issue #743.

### DIFF
--- a/esdoc/manual/formatting.md
+++ b/esdoc/manual/formatting.md
@@ -77,7 +77,7 @@ jsDate.toLocaleDateString('ko-KR', options) // 2018년 4월 28일 토요일 오�
 ### Format patterns
 
 
-Date and time formats are specified by date and time pattern strings using [Java SimpleDateFormat](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) codes.
+Date and time formats are based on the pattern strings from [Java SimpleDateFormat](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html), though js-joda uses more symbols and it interprets a few symbols differently.
 
 | Symbol | Meaning                    | Presentation | Examples                                       |
 | ------ | -------------------------- | ------------ | ---------------------------------------------- |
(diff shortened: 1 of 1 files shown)


pithu:
Hi @janoma, thanks for your PR, give me some time to review it, i am quite busy. 

I wonder why there is a difference between java and js-joda. shouldn't be.

pithu:
Hi @janoma, thanks for looking into this and finding a bug in our documentation. 

But the actual error is that js-joda doc is referencing SimpleDateFormat, thats wrong and pointing to java 7, java 7 does not contain jsr310 yet.

The correct link in js-joda documentation should be https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html. There you find the correct patterns.

And js-joda should always behave in the same way as the jsr310 (all java version >= 8), if not its a bug in js-joda (or in rare cases in upstream java versions)

pithu:
btw: The actual js-joda doc for DateFormatter contains the correct pattern definitions -> https://js-joda.github.io/js-joda/class/packages/core/src/format/DateTimeFormatter.js~DateTimeFormatter.html#static-method-ofPattern

So it looks like the bug is "just" in the manual https://js-joda.github.io/js-joda/manual/formatting.html

janoma:
Thanks @pithu, that makes so much more sense, and the functionality is almost completely implemented in js-joda. I tested locally and updated the documentation accordingly, removing the differences I tried to add before that were relevant to `SimpleDateFormat` only.

There are still the patterns `g` and `v`, which are available in Java 11 but not in js-joda, and also the pattern `B`, available in Java 17 but not in js-joda. I did mention all three in a short paragraph after the table.

Also, I took the liberty of linking to the documentation for Java 17 instead of Java 11, seeing as OpenJDK 11 has recently reached its end of life, and version 17 is the next one with with long term support. I can change it if you feel strongly about that though.

pithu (approved):
Thanks for fixing

PR #768, opened 2025-08-27
Make it less easy to misuse the timezone module

jstasiak:
I spent some time debugging

    ZoneId.of('Europe/Madrid')

crashing with

    DateTimeException: unsupported ZoneId:Europe/Madrid

this morning only to find out that this was actually because the default getRules implementation always threw an exception like this.

I figured I'd try to improve things a bit by changing the error message to be clearer (the problem is not related to any time zone id specifically but it's connected to a missing import) and documenting things more explicitly.


packages/core/src/zone/ZoneRulesProvider.js
```diff
@@ -12,13 +12,14 @@ export class ZoneRulesProvider {
...
+     * This method relies on `@js-joda/timezone` being imported first (for
+     * side effects).
      *
      * @param {string} zoneId
      * @return {ZoneRules}
      */
-    static getRules(zoneId){
-        throw new DateTimeException(`unsupported ZoneId:${zoneId}`);
+    static getRules(_zoneId){
+        throw new DateTimeException(`@js-joda/timezone not imported`);
```
jstasiak, line 22: open question: should this actually throw `DateTimeException` (which the API clients are expected to know about and handle) or maybe something more generic (like plain `Error`) to make it more obvious that this is a sort of a programming error?

jstasiak:
I suspect I need to regenerate the documentation as part of this PR?

pithu:
Hi @jstasiak , thanks for the PR, makes sense. For simplicity i would stay with the DateTimeException

pithu (approved):

PR #792, opened 2026-03-31
Updates to cldr docs and dependency handling

joshkel:
For discussion:

This PR contains two commits:

* Update the [main docs site](https://js-joda.github.io/js-joda/manual/Locale.html#usage) with the @js-joda/locale README's information on how to use the new cldr dependencies.
* Revise cldr dependency handling: By parking the peerDependencies as optional and lazily importing them, this should avoid problems such as https://github.com/js-joda/js-joda/issues/789#issuecomment-4156305343 and https://github.com/js-joda/js-joda/issues/791 without requiring updates to web bundlers' configurations.

I haven't done much testing of this yet; if this approach seems worthwhile, I'd be happy to test further.

I can also split this into two PRs (one for docs, one for dependency handling) if you prefer.

Note that this is technically a break in backward compatibility: Someone who updated to @js-joda/core 6 / @js-joda/locale 5 on Node.js, using NPM as their package manager, would have automatically gotten peerDependencies installed and may have to manually install them after this change.


pithu (commented):
Lets first first discuss a strategy in #793

pithu:
Implemented in another PR

PR #804, opened 2026-06-30
fix: Instant.parse should accept an offset and resolve to the UTC instant (#731)

### LINKED ISSUE #731: Instant.parse() doesn't work with ISO8601 timestamp with offset
Input:
`"@js-joda/core": "^5.6.3"`
```js
Instant.parse("2024-03-07T15:30:00+08:00")
```

## Java (expected)
Instant (`2024-03-07T07:30:00Z`)

## js-joda (actual)
`DateTimeParseException: Text '2024-03-07T15:30:00+08:00' could not be parsed at index 19: 2024-03-07T15:30:00+08:00, at index: 19`

## Workaround

```js
OffsetDateTime.parse("2024-03-07T15:30:00+08:00").toInstant()
```

CedricConday:
Fixes #731

## What was wrong

`Instant.parse` parses with a formatter that ends in a literal `Z`:

```js
.appendValue(SECOND_OF_MINUTE, 2).appendFraction(NANO_OF_SECOND, …).appendLiteral('Z')
```

So an ISO-8601 instant carrying a numeric offset throws instead of parsing:

```js
Instant.parse("2024-03-07T15:30:00+08:00");
// DateTimeParseException: … could not be parsed at index 19
```

`java.time.Instant.parse` (which this library mirrors) accepts an offset and resolves it to the equivalent UTC instant — the expected result here is `2024-03-07T07:30:00Z`.

## Fix

Mirror `java.time`'s `InstantPrinterParser`: append an **offset id** instead of a literal `Z` (`appendOffsetId()` still accepts `Z` as the zero offset), and resolve the local date-time against the parsed `OFFSET_SECONDS` rather than assuming UTC. Printing is unchanged (still emits `Z`).

## Tests

Adds offset cases to the `parse` data provider and a dedicated test for the reported example. Verified:
- `2024-03-07T15:30:00+08:00` → `2024-03-07T07:30:00Z` (was throwing on `master`);
- existing `Z` inputs and all `parseFailures` (bare date-times, partial offsets) still behave as before.

### DIFF
--- a/packages/core/src/format/DateTimeFormatterBuilder.js
+++ b/packages/core/src/format/DateTimeFormatterBuilder.js
@@ -1516,7 +1516,7 @@ class InstantPrinterParser  {
         const parser = new DateTimeFormatterBuilder()
             .append(DateTimeFormatter.ISO_LOCAL_DATE).appendLiteral('T')
             .appendValue(ChronoField.HOUR_OF_DAY, 2).appendLiteral(':').appendValue(ChronoField.MINUTE_OF_HOUR, 2).appendLiteral(':')
-            .appendValue(ChronoField.SECOND_OF_MINUTE, 2).appendFraction(ChronoField.NANO_OF_SECOND, minDigits, maxDigits, true).appendLiteral('Z')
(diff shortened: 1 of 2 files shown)


pithu (approved):
Thanks for contributing

PR #805, opened 2026-06-30
fix(core): correct Duration.toString for negative seconds on minute boundary

spokodev:
## Problem

`Duration.toString()` produces a malformed, sign-inconsistent string for negative durations when `seconds < 0`, `nanos > 0`, and `seconds` is an exact multiple of 60 (a whole-minute boundary).

The current implementation derives `hours`/`minutes`/`secs` straight from the raw `this._seconds` field and gates its sign branches on the post-division `secs` value. When `seconds % 60 === 0`, the unadjusted `secs` is `0`, so the borrow across the minute/hour boundary is skipped and the output mixes a negative hour/minute term with a positive seconds term.

`Duration.toString` is documented to produce the canonical normalized ISO-8601 form in which every term carries the same sign (the long-standing `java.time.Duration` contract; the same defect is tracked upstream as JDK-8146278). The `PT-1H0.5S` style output (negative hour, positive seconds) is not a valid value of that form and does not round-trip cleanly.

Reproduced on `@js-joda/core@6.0.1` (current `main` has the same code); the expected column is the documented canonical form:

| input | js-joda | expected |
|---|---|---|
| `Duration.ofSeconds(-3600, 500000000)` | `PT-1H0.5S` | `PT-59M-59.5S` |
| `Duration.ofSeconds(-60, 500000000)` | `PT-1M0.5S` | `PT-59.5S` |
| `Duration.ofSeconds(-86400, 500000000)` | `PT-24H0.5S` | `PT-23H-59M-59.5S` |
| `Duration.ofSeconds(-120, 1)` | `PT-2M0.000000001S` | `PT-1M-59.999999999S` |

## Root cause

In `packages/core/src/Duration.js`, `toString()`:

1. computes hours/minutes/secs from the raw `this._seconds` instead of a value that accounts for the sub-second offset when `seconds < 0 && nanos > 0`;
2. gates the seconds-sign branch on the post-division `secs` (`secs < 0` / `secs === -1`) rather than the duration's actual sign;
3. gates the nano formatting on `secs < 0` rather than the duration's actual sign.

When `seconds % 60 === 0` the unadjusted `secs` is `0`, so the sign of a negative sub-second duration is lost across the minute/hour boundary.

## Fix

`toString` is reimplemented in js-joda's own idioms. js-joda stores a negative duration as a negative second count plus a non-negative nano offset (e.g. `-0.5s` is `seconds = -1, nanos = 500000000`). The fix recovers, from that representation, the signed whole-second part and the magnitude of the sub-second fraction separately:

- when the duration is negative and has a nano offset, one second of the offset belongs to the whole part (`wholeSeconds = seconds + 1`) and the remaining fraction is its complement against a full second (`NANOS_PER_SECOND - nanos`);
- hours/minutes/secs come from `wholeSeconds`;
- the seconds field gets an explicit `-0` when the whole part is zero but the duration is a negative fraction;
- the fraction is printed as a zero-padded nine-digit value with trailing zeros trimmed.

`Duration.parse` and the arithmetic are untouched; only `toString` changes. The correct outputs were determined from the documented `Duration.toString` behavior (JDK-8146278); no OpenJDK source was copied or translated.

## Tests

`data_toString` in `packages/core/test/reference/DurationTest.js` previously omitted the negative-seconds-with-non-zero-nanos rows. This PR adds them: the four cases above plus the sub-second cases where the whole-second part is zero (`ofSeconds(-1, 5e8)` -> `PT-0.5S`, `ofSeconds(-61, 5e8)` -> `PT-1M-0.5S`, `ofSeconds(-3601, 5e8)` -> `PT-1H-0.5S`), asserting the documented output.

Before the fix these rows fail (e.g. `expected 'PT-1M0.5S' to deeply equal 'PT-59.5S'`). After the fix `DurationTest` is 154/154 and the full `@js-joda/core` mocha suite passes (2585 passing, 0 failing), including all pre-existing `toString` cases and the previously-correct negative cases such as `ofSeconds(-30, 5e8)` -> `PT-29.5S` and `ofSeconds(-3600, 0)` -> `PT-1H`.

### DIFF
--- a/packages/core/src/Duration.js
+++ b/packages/core/src/Duration.js
@@ -1209,9 +1209,22 @@ export class Duration extends TemporalAmount /*implements TemporalAmount, Compar
         if (this === Duration.ZERO) {
             return 'PT0S';
         }
-        const hours = MathUtil.intDiv(this._seconds, LocalTime.SECONDS_PER_HOUR);
-        const minutes = MathUtil.intDiv(MathUtil.intMod(this._seconds, LocalTime.SECONDS_PER_HOUR), LocalTime.SECONDS_PER_MINUTE);
-        const secs = MathUtil.intMod(this._seconds, LocalTime.SECONDS_PER_MINUTE);
+        // A negative duration is stored as a negative second count plus a
+        // non-negative nano offset, e.g. -0.5s is held as seconds = -1,
+        // nanos = 500000000. To render ISO-8601 components that all carry the
+        // duration's sign, recover the signed whole-second part and the
+        // magnitude of the sub-second fraction separately: when the duration is
+        // negative and has a fractional offset, one second of that offset
+        // belongs to the whole part, and the remaining fraction is its
+        // complement against a full second.
+        const negativeFraction = this._seconds < 0 && this._nanos > 0;
+        const wholeSeconds = negativeFraction ? this._seconds + 1 : this._seconds;
+        const fractionNanos = negativeFraction ? LocalTime.NANOS_PER_SECOND - this._nanos : this._nanos;
+
+        const hours = MathUtil.intDiv(wholeSeconds, LocalTime.SECONDS_PER_HOUR);
+        const minutes = MathUtil.intDiv(MathUtil.intMod(wholeSeconds, LocalTime.SECONDS_PER_HOUR), LocalTime.SECONDS_PER_MINUTE);
+        const secs = MathUtil.intMod(wholeSeconds, LocalTime.SECONDS_PER_MINUTE);
+
         let rval = 'PT';
         if (hours !== 0) {
             rval += `${hours}H`;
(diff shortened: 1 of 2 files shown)


pithu:
Thanks for this! 🙏 The bug analysis is spot on and the fix is correct — I traced the new cases and the existing negative ones, no regressions.

One thing before merging: **provenance.** js-joda is BSD and deliberately ports from ThreeTen-Backport (also BSD), not from OpenJDK's `java.time` (GPL). Current threeten-bp still has this bug, so there was no BSD source to port from — which I imagine is why you looked at the JDK.

So, just to keep things clean: did you translate this from OpenJDK source, or write it yourself using the JDK only to figure out the correct output? If the latter (your own code, JDK behavior/JDK-8146278 as a spec), we're all set. If any OpenJDK source was copied or translated, the fix would need to be reworked as an independent implementation before we can merge it — we can't take GPL-derived code into the BSD codebase regardless of how the PR is described.

This wasn't clearly documented before, so that's on us — we're adding a note to the README to make the rule explicit. Thanks again! 🙌

spokodev:
Thanks for the careful review and for spelling out the provenance rule. Straight answer: the first version was written from the JDK's algorithm (I had been treating java.time's `toString` as the reference), so by your rule it counts as JDK-derived. I have reworked it.

The version now on the branch is an independent implementation: it derives from js-joda's existing `toString` plus JDK-8146278 / the documented `java.time` `toString` contract used only as the spec for the correct output. It does not copy or translate any OpenJDK source. It differs from the JDK in how it renders the fraction: it recovers the signed whole-second part and the sub-second magnitude (`NANOS_PER_SECOND - nanos`) and prints a zero-padded, trailing-trimmed fraction, rather than the JDK's prepend-and-strip technique. The borrow-a-second step is the arithmetic the negative sub-second case requires.

Verified: `DurationTest` 154/154 and the full core suite 2585/2585 pass, and the outputs match the documented behavior across the boundary cases. I also reworded the PR to drop the "port OpenJDK" framing and reference JDK-8146278 as the behavioral spec. Let me know if you need anything else for BSD-cleanliness.

pithu:
Reviewed the reworked version — thanks for turning this around so quickly! 🙌

The implementation is now clearly independent (complement-based fraction + zero-pad/trim, no OpenJDK structure), and the description accurately frames it as derived from documented behavior with no source copied. Provenance concern resolved. 👍

I re-traced all seven new cases plus the regressions and everything checks out. LGTM.

pithu (approved):

PR #807, opened 2026-07-02
fix: parse zone ids that share a prefix with a fixed id

### LINKED ISSUE #681: ZonedDateTime with Zulu or GMT0 timezone is unparsable from string?
This seems like a bug? js-joda is quite happy to construct a ZonedDateTime with the Zulu or GMT0 timezones and convert it to a string but unable to re-parse it from the same string output?

```typescript
import {ZonedDateTime, LocalDateTime, ZoneId} from '@js-joda/core'
import '@js-joda/timezone'

console.log('start')
const now = LocalDateTime.now()
for(const zoneId of ZoneId.getAvailableZoneIds()) {
  const date = ZonedDateTime.of(now, ZoneId.of(zoneId))
  const dateString = date.toString()
  try {
    ZonedDateTime.parse(dateString)
  } catch(e) {
    console.error(e)
  }
}
console.log('end')
```

Output:
```
start 
DateTimeParseException: Text '2023-04-17T14:25:43.635Z[GMT0]' could not be parsed, unparsed text found at index 24: 2023-04-17T14:25:43.635Z[GMT0], at index: 24
DateTimeParseException: Text '2023-04-17T14:25:43.635Z[Zulu]' could not be parsed, unparsed text found at index 24: 2023-04-17T14:25:43.635Z[Zulu], at index: 24
end 
```

apoorva-01:
Fixes #681.

js-joda prints a `ZonedDateTime` for `GMT0` or `Zulu` as `…Z[GMT0]` / `…Z[Zulu]`, then throws parsing it back. The parser matches the `GMT`/`UT`/`UTC` prefixes and a leading `Z` before it checks the region table and stops on the first hit, so `GMT0` stops at `GMT` and `Zulu` at `Z`. Both are real zone ids longer than the prefix, so now the longest match wins.

One call for you: longest-wins, not always-prefer-region. `GMT`/`UTC` are real zone ids too, so region-first would break `GMT+03:30` / `UTC+01:30`. Happy to change it.

Tests in `ZoneIdPrinterParserTest.js`; core suite green, eslint clean, nothing from OpenJDK.

### DIFF
--- a/packages/core/src/format/parser/ZoneIdPrinterParser.js
+++ b/packages/core/src/format/parser/ZoneIdPrinterParser.js
@@ -71,7 +71,67 @@ export class ZoneIdPrinterParser {
             return ~position;
         }
 
-        // handle fixed time-zone IDs
+        // A fixed ID (Z, UT, UTC, GMT, an offset) can also be the start of a
+        // longer region id such as GMT0 or Zulu, so parse both and keep the
+        // longest match (as the class doc above promises).
+        const region = this._parseRegion(text, position);
+        const fixedEndPos = this._parseFixedId(context, text, position);
+
+        const fixedLength = fixedEndPos >= 0 ? fixedEndPos - position : 0;
+        if (region != null && region.parseLength > fixedLength) {
+            context.setParsedZone(ZoneRegion.ofId(region.parsedZoneId));
+            return position + region.parseLength;
+        }
+        return fixedEndPos;
+    }
+
+    /**
+     * Finds the longest available zone region id matching text at position,
+     * without touching the parse context.
+     *
+     * @param {String} text
+     * @param {number} position
+     * @return {?{parsedZoneId: string, parseLength: number}}
+     */
+    _parseRegion(text, position) {
+        const availableZoneIds = ZoneRulesProvider.getAvailableZoneIds();
+        if (availableZoneIds.length === 0) {
+            return null;
+        }
+        if (zoneIdTree.size !== availableZoneIds.length) {
+            zoneIdTree = ZoneIdTree.createTreeMap(availableZoneIds);
+        }
+
+        const maxParseLength = text.length - position;
+        let treeMap = zoneIdTree.treeMap;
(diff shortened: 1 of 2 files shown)


pithu:
Nice fix — thanks for the clear write-up in the description, especially calling out the longest-wins vs. region-first tradeoff. That's the right call: `GMT`/`UTC` are real zone ids, so region-first would break `GMT+03:30`. Longest-match-wins is also consistent with the existing class doc ("looks for the longest matching string").

The tests are the strong part here — `data_longestMatch` pins down exactly the cases that matter (`GMT0`/`Zulu` winning, `GMT+03:30`/`UTC+01:30` offsets still beating the bare region, and the `GMT`/`Z0`/`UTC0`/`GMTZ` ties where the fixed id wins). 👍

One **optional, non-blocking** suggestion if you feel like it:

Instead of threading `_preferRegionIfLonger(...)` through the four fixed-ID branches, the "longest wins" decision could live in a single place. Extract the fixed-ID handling into one `_parseFixedId(...)` that just returns an end position (or `~position`), then compare once at the tail:

```js
const region = this._parseRegion(text, position);
const fixedEndPos = this._parseFixedId(context, text, position);

const fixedLength = fixedEndPos >= 0 ? fixedEndPos - position : 0;
if (region != null && region.parseLength > fixedLength) {
    context.setParsedZone(ZoneRegion.ofId(region.parsedZoneId));
    return position + region.parseLength;
}
return fixedEndPos;
```

That lets the UT/UTC/GMT/Z branches go back to plain `return`s and removes `_preferRegionIfLonger` entirely — same behavior, one decision point. Purely a readability preference though; the current version looks correct.

pithu (approved):

apoorva-01:
Good call, that reads better. Pushed it: the fixed-id handling moved into `_parseFixedId` and the longest-wins comparison happens once at the tail, so `_preferRegionIfLonger` is gone and the branches are plain returns again.

pithu:
Thanks for contributing @apoorva-01

PR #808, opened 2026-07-08
fix(core): truncate Instant toward negative infinity for pre-epoch values

spokodev:
`Instant.truncatedTo` divided the nano-of-day by the unit length with integer division that rounds toward zero. For instants before 1970 the nano-of-day is negative, so truncation rounded up toward the epoch instead of down. For example `1969-12-31T23:59:30Z` truncated to minutes returned `1970-01-01T00:00:00Z` instead of `1969-12-31T23:59:00Z`.

Using `floorDiv` floors toward negative infinity, matching `java.time.Instant.truncatedTo` and the fix in ThreeTen PR 54. The two reference assertions that were disabled for this are re-enabled.


pithu (approved):

pithu:
thanks for contributing @spokodev

PR #809, opened 2026-07-08
fix(core): initialize parsed excessDays to Period.ZERO

spokodev:
`DateTimeFormatter.parsedExcessDays()` returned `undefined` after parsing an input without excess days. Following the documented pattern of querying the parse result and inspecting the `Period` then threw instead of yielding a zero period.

`java.time` initializes the parsed result's `excessDays` to `Period.ZERO`. This change does the same, so the query returns `Period.ZERO` when nothing overflowed and still returns the real excess (for example `24:00` giving `P1D`) when present. The four reference assertions that check the zero case are re-enabled.


pithu (approved):

pithu:
thanks for contributing @spokodev

Issue #198, opened 2017-11-03
Some Zone Ids do not work


landrys:
console.log(ZonedDateTime.now(ZoneId.of('America/New_York')).toString());
gives 
throw new _errors.DateTimeException('unsupported ZoneId:' + zoneId);

phueper:
Hi @landrys,

Thanks for reporting this and thanks for using js-joda. 

It seems you have tried using timezones just with "plain" js-joda. We have split the timezone functionality in a separate module: https://github.com/js-joda/js-joda-timezone

Could you please verify that your problem still exists when using js-joda-timezone as described in its README, and if it does exist, report it in the js-joda-timezone github repository?

I will close this issue here, but please feel free to reopen or create there. 

Regards, Pattrick

landrys:
All good now. Thanks for information.

PR #810, opened 2026-07-09
fix(locale): exclude non-uniquely-parsable text styles from parsing

CedricConday:
Ports the upstream threetenbp fix for issue #198 (`SimpleDateTimeTextProvider.LocaleStore`) to `@js-joda/locale`.

### Problem
`LocaleStore` keeps a text style even when two field values map to the same text — e.g. narrow month text where "J" is January, June *and* July. It builds a first-wins reverse-parse map and still exposes the style as parsable, so parsing ambiguous text silently resolves to whichever value was inserted first. This diverges from the reference implementation.

### Fix
Mirror threetenbp #198: when any text within a style collides, skip the **entire** style so it is not parsable — `getTextIterator(style)` (and the combined all-styles list) return `null`. This matches the `continue outer` behaviour in the Java `SimpleDateTimeTextProvider.LocaleStore`.

### Test
Added `LocaleStoreTest.js`: a style with duplicate text ("J" → 1/6/7) is not parsable; a style with unique text stays parsable with all entries present. Verified the test fails on the pre-fix code and passes after.

Upstream reference: `ThreeTen/threetenbp@d3fb4cd5b` — "Fix loop in SimpleDateTimeTextProvider (#198)".

### DIFF
--- a/packages/locale/src/format/LocaleStore.js
+++ b/packages/locale/src/format/LocaleStore.js
@@ -48,13 +48,25 @@ export class LocaleStore {
         Object.keys(valueTextMap).forEach((style) => {
             const reverse = {};
             const list = [];
+            let parsable = true;
             Object.keys(valueTextMap[style]).forEach((key) => {
-                const value = valueTextMap[style][key];
-                if (reverse[value] === undefined) {
-                    reverse[value] = createEntry(value, parseInt(key));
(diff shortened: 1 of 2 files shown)


pithu (approved):

pithu:
thanks for contributing @CedricConday

PR #812, opened 2026-07-21
fix(core): correct ISO_WEEK_DATE and enable ISO week/quarter parsing

binggao1230:
`DateTimeFormatter.ISO_WEEK_DATE` returns the wrong date in both directions for most years, silently. It is built on `ChronoField.YEAR` + `ALIGNED_WEEK_OF_YEAR` (a week counter that resets on Jan 1) instead of the ISO week fields, so any year whose Jan 1 is not a Monday is off by a week.

```js
LocalDate.parse('2014-W01-1', DateTimeFormatter.ISO_WEEK_DATE).toString();
// '2014-01-06', should be '2013-12-30'
LocalDate.of(2013, 12, 30).format(DateTimeFormatter.ISO_WEEK_DATE);
// '2013-W52-1', should be '2014-W01-1'
```

Both `java.time` (the reference API) and Python's `date.fromisocalendar` give `2013-12-30` / `2014-W01-1`. The lone existing `ISO_WEEK_DATE` test used `2018-W17-6`, and 2018 is one of the rare years where Jan 1 is a Monday, so the aligned-week counter coincides with the ISO week and the bug stayed hidden.

The underlying cause is that `DateTimeBuilder.resolve()` never ran the per-field resolve step, so `IsoFields.WEEK_OF_WEEK_BASED_YEAR.resolve()` and `IsoFields.DAY_OF_QUARTER.resolve()` were never called during parsing. The correct ISO week fields could not be parsed at all, which is why the formatter fell back to the wrong-but-parseable `ChronoField.YEAR` + `ALIGNED_WEEK_OF_YEAR`.

This invokes those fields' `resolve()` from `DateTimeBuilder.resolve()` and defines `ISO_WEEK_DATE` with `WEEK_BASED_YEAR` + `WEEK_OF_WEEK_BASED_YEAR`, matching the java.time definition. As a side effect, quarter dates now parse too: a `YEAR` + `QUARTER_OF_YEAR` + `DAY_OF_QUARTER` formatter previously threw `DateTimeParseException` and now round-trips.

Note on scope: js-joda keeps parsed values in an `EnumMap` keyed by field name, so java.time's generic resolve loop over field objects isn't directly portable without reworking that map. Instead the two ISO fields that define a `resolve()` are handled explicitly, which covers the whole ISO week/quarter class in core. Locale `WeekFields` parsing has the same underlying limitation and could be addressed separately.

Tests: the previously disabled `TestIsoFields` week-parse test (`/* TODO weekday parser`) is restored, and week/quarter parse+format+round-trip coverage is added for non-Monday-Jan-1 years, single-digit weeks, and 53-week years. The full `@js-joda/core` suite passes (2595) as does `@js-joda/extra` (719).

### DIFF
--- a/packages/core/src/format/DateTimeBuilder.js
+++ b/packages/core/src/format/DateTimeBuilder.js
@@ -14,6 +14,7 @@ import { ResolverStyle } from './ResolverStyle';
 import { IsoChronology } from '../chrono/IsoChronology';
 import { ChronoLocalDate } from '../chrono/ChronoLocalDate';
 import { ChronoField } from '../temporal/ChronoField';
+import { IsoFields } from '../temporal/IsoFields';
 import { TemporalAccessor } from '../temporal/TemporalAccessor';
 import { TemporalQueries } from '../temporal/TemporalQueries';
 
@@ -148,11 +149,10 @@ export class DateTimeBuilder extends TemporalAccessor {
         // this._mergeInstantFields();
         this._mergeDate(resolverStyle);
         this._mergeTime(resolverStyle);
-        //if (resolveFields(resolverStyle)) {
-        //    mergeInstantFields();
-        //    mergeDate(resolverStyle);
-        //    mergeTime(resolverStyle);
-        //}
+        if (this._resolveFields(resolverStyle)) {
+            this._mergeDate(resolverStyle);
+            this._mergeTime(resolverStyle);
+        }
         this._resolveTimeInferZeroes(resolverStyle);
         //this._crossCheck();
         if (this.excessDays != null && this.excessDays.isZero() === false && this.date != null && this.time != null) {
@@ -162,6 +162,30 @@ export class DateTimeBuilder extends TemporalAccessor {
         //resolveFractional();
         this._resolveInstant();
         return this;
+    }
+
+    /**
+     * Resolves the date fields that are not {@link ChronoField}s by delegating to their own
+     * `resolve()`, the step java.time performs generically. As parsed values are held in a
+     * name-keyed map, the ISO fields that define a `resolve()` (week-of-week-based-year and
+     * day-of-quarter) are handled here. Returns whether a date was produced.
+     *
+     * @param {ResolverStyle} resolverStyle
+     * @return {boolean}
+     * @private
+     */
+    _resolveFields(resolverStyle) {
+        let resolved = false;
+        for (const field of [IsoFields.WEEK_OF_WEEK_BASED_YEAR, IsoFields.DAY_OF_QUARTER]) {
+            if (this.fieldValues.containsKey(field)) {
+                const date = field.resolve(this.fieldValues, this, resolverStyle);
+                if (date != null) {
+                    this._checkDate(date);
+                    resolved = true;
+                }
+            }
+        }
+        return resolved;
     }
 
     /**
--- a/packages/core/src/format/DateTimeFormatter.js
+++ b/packages/core/src/format/DateTimeFormatter.js
@@ -21,6 +21,7 @@ import { ResolverStyle } from './ResolverStyle';
 
 import { IsoChronology } from '../chrono/IsoChronology';
 import { ChronoField } from '../temporal/ChronoField';
+import { IsoFields } from '../temporal/IsoFields';
 import { createTemporalQuery } from '../temporal/TemporalQuery';
 
 /**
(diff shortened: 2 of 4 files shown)


pithu:
Thanks for this — the core fix looks right. I checked a 12,000-day format→parse round trip through `ISO_WEEK_DATE` with zero mismatches, plus 53-week years, negative/`EXCEEDS_PAD` years, `LocalDate.MIN`/`MAX`, and strict rejection of week 53 in a 52-week year and of day-of-week 8. Both suites pass here too.

One thing I'd like changed before merge, in `DateTimeBuilder.resolve()`: `_resolveFields()` runs *after* the first `_mergeDate()`/`_mergeTime()` pair, whereas upstream runs the per-field resolve step *before* date merging. Two silent misbehaviours follow from the order:

**1. Conflicting date fields go undetected.** With a formatter combining `YEAR` + `MONTH_OF_YEAR` + `DAY_OF_MONTH` and `WEEK_BASED_YEAR` + `WEEK_OF_WEEK_BASED_YEAR` + `DAY_OF_WEEK`, parsing `'2020-06-01 2020-W01-1'` returns `2019-12-30`. `_mergeDate` builds `2020-06-01` first; its cross-check passes because the only shared `ChronoField` (`DAY_OF_WEEK` = 1) happens to agree; then `_resolveFields` → `_checkDate` → `_addObject` overwrites `this.date` without comparing it against the already-resolved date. Same with `YEAR` + `DAY_OF_YEAR`: `'2020-153 2020-W01-1'` → `2019-12-30`. Reference behaviour is a `Conflict found: 2019-12-30 differs from 2020-06-01`. The hole only opens when the cross-checked `ChronoField`s coincide — a disagreeing day-of-week does still throw.

**2. Quarter fields are silently dropped** once `YEAR` has been consumed. `IsoChronology.resolveDate` removes `YEAR` during `_mergeDate`, so `DAY_OF_QUARTER.resolve()` then reads `fieldValues.get(ChronoField.YEAR) == null` and returns `null`. A `YEAR` + `MONTH_OF_YEAR` + `DAY_OF_MONTH` + `QUARTER_OF_YEAR` + `DAY_OF_QUARTER` formatter parsing `'2020-06-01 1-05'` returns `2020-06-01`, ignoring the (contradictory) quarter fields rather than raising a conflict.

Both are reachable only because of the new ordering, so it's this PR's code rather than extra scope. The fix is to move the resolve step ahead of the merges:

```js
if (this._resolveFields(resolverStyle)) {
    // this._mergeInstantFields();
}
this._mergeDate(resolverStyle);
this._mergeTime(resolverStyle);
```

As belt-and-braces, having `_checkDate`/`_addObject` reject a second, differing date instead of overwriting would close the hole independently of ordering — but the reorder alone is the faithful fix.

Could you add a test for each case: `'2020-06-01 2020-W01-1'` should throw a conflict, and the quarter-fields combination should not silently drop the quarter.

Two non-blocking notes: `_resolveFields` returns "a date was produced" where the reference returns "something changed" and loops until stable — harmless with only these two fields, but worth a comment; and the second `_mergeTime` pass is a no-op in practice, since `EnumMap.containsKey` treats removed (undefined) entries as absent.

binggao1230:
Addressed in `7fe65978`. `_resolveFields()` now runs before date/time merging, so existing calendar fields are cross-checked instead of overwritten or dropped.

I added both requested conflict regressions, plus matching calendar/week and calendar/quarter controls verified against `java.time`. The old-order mutant fails both conflict tests; an over-strict second-date mutant fails the matching-week control.

Validation: core 2,597 passing / 8 pending; extra 719 passing; lint clean; TypeScript definitions 22 passing; `build-dist` passes.

pithu:
The reorder looks right — thanks. I re-ran the cases against the PR head and they all behave now:

| case | before | now |
|---|---|---|
| `'2020-06-01 2020-W01-1'` | `2019-12-30` | throws `Conflict found: Field Year 2019 differs from Year 2020` |
| `'2018-04-28 2019-W17-6'` | `2019-04-27` | throws conflict |
| `'2020-153 2020-W01-1'` (day-of-year) | `2019-12-30` | throws conflict |
| `'2018-2-15 05-15'` (quarter) | silently `2018-05-15` | throws `Conflict found: Field MonthOfYear 4 differs from MonthOfYear 5` |
| `'2018-2-15 04-15'` (quarter, agreeing) | — | `2018-04-15` |
| `'2018-2-15'` (quarter alone) | — | `2018-04-15` |

Quarter fields are no longer dropped, since `DAY_OF_QUARTER.resolve()` now sees `YEAR` before `_mergeDate` consumes it. Full core suite: 2597 passing, 8 pending.

**Optional, not blocking this PR** — threeten-bp's `ISO_WEEK_DATE` ends with an optional offset before `toFormatter`:

```js
    .appendValue(ChronoField.DAY_OF_WEEK, 1)
    .optionalStart()
    .appendOffsetId()
    .toFormatter(ResolverStyle.STRICT).withChronology(IsoChronology.INSTANCE);
```

Without it, `ISO_WEEK_DATE` can't parse `'2018-W17-6+01:00'` and drops the offset when formatting an `OffsetDateTime`/`ZonedDateTime`. Not a regression — the old definition lacked it too — but since this PR brings the definition in line with the reference, it's cheap to include. Everything else in the new builder chain matches threeten-bp exactly. Happy either way: here, or as a follow-up.

